Issue #50, opened 2022-08-16
Create encoding "guidelines"


ahankinson:
While the version 2 specification will help clarify the boundaries of the Plaine & Easie code, it will also create a hole. Previous versions were part specification, part encoding guideline. In Version 2 the guidelines are largely going to feature less in the specification, but there is likely still an audience for them.

So we should create a second document that can contain implementation guides, best practices, or other guidelines. Like "record transposing instruments in a non-transposed way" or "Should contain 3-4 measures", etc.

PR #53, opened 2022-10-07
Fixed: Move principles to guidelines

ahankinson:
This commit moves the "General Principles" section to the encoding guidelines.

Refs #50

### DIFF
--- a/v2/guidelines.html
+++ b/v2/guidelines.html
@@ -70,5 +70,25 @@
 <section id="abstract">
     <p></p>
 </section>
+<section class="informative">
+    <h2>General Principles</h2>
+    <div class="issue" data-number="30"></div>
+    <p>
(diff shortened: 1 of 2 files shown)


v2/guidelines.html
```diff
@@ -70,5 +70,25 @@
 <section id="abstract">
     <p></p>
 </section>
+<section class="informative">
+    <h2>General Principles</h2>
+    <div class="issue" data-number="30"></div>
+    <p>
+        Music incipits help identify works and facilitate the comparison of historical musical sources.
+    </p>
```
lpugin, line 78: I think we would still need to have an introduction paragraph at the beginning of the specs

ahankinson: Yes, I agree -- but we can cover this in a different issue, #57

lpugin (approved):

jenniferward (approved):

Issue #35, opened 2022-08-15
Move field references for clefs, keysig, timesig, etc. to single string representation section.


ahankinson:
In some places in the PAE specification the value for the single-string representation is given, e.g., for key signature it says

> Begin this field with the character <code>$</code>

However, this is only necessary when using the "single-string" representation. All other representations, including MARC, do not require this.

The exception to this is when discussing clef/keysig/timesig changes, so references to the field delimiters will be kept in that section.

PR #54, opened 2022-10-07
Fixed: Rewrite key signature section

### LINKED ISSUE #15: Order of sharps or flats in key signature
We should decide if we want ordered sharps/flats in the key signature or allow arbitrary sequences. For example: do we tolerate D major written as $xCF or it must be $xFC?

I found 225 cases with wrong order of sharps/flats (including cases like $Bb), see [attached list](https://github.com/rism-digital/pae-code-spec/files/7337982/keysig_wrong_notes_order.csv). The real number of those could be slightly higher than this, since grammatically wrong key signatures are excluded from the analysis.

The actual paec-grammar version does not check the order of sharps/flats.

### LINKED ISSUE #16: Repeated sharps or flats in key signature
I found 76 cases with repeated sharps/flats in the key signature (see [list](https://github.com/rism-digital/pae-code-spec/files/7339684/keysig_multiple_notes.csv), the real case number might be higher). Examples are: $bBEE, $xFCGF, but also $xFCG[G]

Should we correct those entries and modify the paec-grammar in order to detect them?

What about $xFCG[G] ? We probably have no chance to know which G is too much

### LINKED ISSUE #17: Clarify use of square brackets with more accidentals in key signature
The spec should explicitly state that key signatures with consecutive accidentals requiring square brackets must be included within a single square bracket. Consecutive square brackets are not allowed. For example: $xF[CG] is allowed, $x[F]C[G] is allowed, but $xF[C][G] is forbidden.

### LINKED ISSUE #26: Clarify whether `n` is a valid accidental type in key signatures.
Not sure if this is a Muscat thing, a Verovio thing, or a P&E thing, so I'll start discussion here.
The guidelines say: 
```
12. Change of clef, key or time signature:
Use % to change the clef, $ to change the key, and @ to change the time signature. Follow this with the new indication (clef, key, or time), followed by a space.

Examples:
%C-1 '2A
%C-1 $xFC '8B
@3/2 '1C
$nBE $xFC
```

I'm interested in the last example. With the new Verovio validation, if you enter something like `4CDEF/$nBE $xFC AB''C` then it's not happy because you can't change the key signature more than once in a measure: 
image

But adding the required bar line `4CDEF/$nBE /$xFC AB''C` isn't very satisfying optically:
image

But maybe you can just leave out the cancelled key signature and go directly to the new one:
`4CDEF/$xFC AB''C`
image

The previous Verovio version (still visible on muscat-training) didn't print the cancelled key signature anyways, so I guess we haven't missed this feature. I am wondering if Verovio should be adjusted, or if we can make a shortcut in the guidelines to go directly to the new key signature (analogue to going directly to the new clef or meter).

### LINKED ISSUE #36: Make key signature a recommended field
Currently the key signature can be empty. However, this makes it unclear whether the key signature is missing, or whether it is a completely natural key signature.

The section should be clarified to say that encoders **should** provide a value, and that a value of `n` can be given if no key signature is present.

It should also clarify that parsers should assume a value of `n` if no key signature is present.

Split off from #11 
Related to #26

ahankinson:
Reworks the key signature section to the new language. Fixes several issues. Adds three new examples to the list of example key signatures.

Fixes #15
Fixes #16
Fixes #17
Fixes #26
Refs #35
Fixes #36

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -185,9 +185,6 @@
     </section>
     <section>
         <h3>Key Signature</h3>
-        <aside class="note">
-            UNIMARC field 036 $n; MARC21 field 031 $n
-        </aside>
         <p>
             Encodings SHOULD provide a key signature field. If no key signature is present, a value of <code>n</code>
             is used to indicate a natural key signature. A missing key signature SHOULD be interpreted as a natural


lpugin (approved):
Looks good. The only question I have is about key signature changes and cancellation. I think we need to decide if `n` MAY be given (for example when changing from `xFCG` to `nxF`. If yes, we also need to decide the allowed order(s).

Please ignore my question if this is going to be treated separately.

ahankinson:
@lpugin I think we can treat that separately, since that would only occur when dealing with inline key changes (you would never have a key change in the key signature field). So we can address that when we get to it. 

I've created #58 to make sure we address this later.

ahankinson:
@BaMikusi @jenniferward could you please review this and merge or provide feedback? Thanks!

ahankinson:
Replaced by #67

PR #55, opened 2022-10-07
Fixed: Rewrite time signature section

### LINKED ISSUE #37: Make time signature an optional field
Time signatures should be an optional field, since they are not necessary.

Split off from #11

ahankinson:
Rewrites the time signature section to the new format. Tidied up the examples a bit.

Fixes #37
Refs #35


jenniferward (approved):
I approve but for some reason the column heading called "Note" threw me off at first because I was expecting musical notes, not prose. Can we change it? Maybe Remarks, Comments, Commentary?

ahankinson:
Sure, but we should probably do that across all the examples (there are others) so I will open an issue.

#59

ahankinson:
@lpugin @BaMikusi could you please review this and merge / provide feedback? Thanks!

v2/index.html
```diff
@@ -247,16 +247,27 @@ <h3>Key Signature</h3>
...
+            "major" prolation; no <code>.</code> indicates "minor" prolation. Numbers following these signs
+            indicate <em>Modus cum tempore</em> signs with a numeral. A <code>/</code> character may follow the
+            <em>tempus</em> character to indicate diminution.
         </p>
-        <div class="issue" data-number="37"></div>
-        <div class="issue" data-number="35"></div>
         <p>
-            Fractional or numeric values are transcribed as fractions and mensuration signs are transcribed with a
-            lowercase letter, if necessary followed by <code>/</code> or <code>.</code>:
+            CWMN time signatures that indicate alternating measures MAY be provided by transcribing both. These MUST
```
lpugin, line 267: The is not problem with the description here. However, I wonder if we could avoid in PAE2 the use of a space. Ideally, a space should indicate the end of the time signature. In fact, I think we have a real problem if we want to forbid spaces in one-line incipit input, where the score definition is delimited by a space.

A possible character we could use insteas is `-`. Any other ideas?

ahankinson: The pipe `|`, although I am conscious that some keyboard layouts might not have this readily accessible?

ahankinson: We can create a new issue to track this change so that it doesn't hold up this PR

lpugin:
I think we need to allow for `c3/2` to be a valid time signature, which can be either CMWN or mensural, I think

ahankinson:
image

Verovio ignores the 3/2 for CWMN at the moment, so I'm not sure this is a valid time sig? I'm not even sure what that would mean?

lpugin:
Yes, Verovio needs to be [adjusted](https://github.com/rism-digital/verovio/issues/2574). There has been a long-standing discussion in MEI about it too, but I think it is fairly common.

PR #63, opened 2022-12-06
Fixed: Clarify structure of single-line PAE

### LINKED ISSUE #62: Order and spacing for score definition elements in single line input
In v1, single line input is not strict about the order of clef, keysig and timesig. We can also see data with a space between them or not. For example
```
%C-2$xF@c/ ...
```
or
```
%C-2 @c/ $xF ...
```

In v2, we should have a fix order and forbid spacing. The order should be clef, keysig, timesig because this is the order they appear in a score. This means that only
```
%C-2$xF@/c ...
```
Would be valid

ahankinson:
Makes the single-line text normative.

Fixes #62


lpugin (approved):

ahankinson:
Fixed the merge conflict.

lpugin (approved):

PR #67, opened 2024-05-15
Updated key signature

### LINKED ISSUE #15: Order of sharps or flats in key signature

### LINKED ISSUE #16: Repeated sharps or flats in key signature

### LINKED ISSUE #17: Clarify use of square brackets with more accidentals in key signature

### LINKED ISSUE #26: Clarify whether `n` is a valid accidental type in key signatures.

### LINKED ISSUE #36: Make key signature a recommended field

ahankinson:
Replaces #54 

Fixes #15
Fixes #16
Fixes #17
Fixes #26
Refs #35
Fixes #36

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -90,7 +90,15 @@
         Adapted from the RISM guidelines.
     </p>
 </section>
-
+<section id="terminology">
+    <h2>Terminology</h2>
+    <dl>
+        <dt><dfn>Circle of fifths</dfn></dt>
+        <dd><!-- TODO --></dd>
+        <dt><dfn>Transcriber</dfn></dt>
+        <dd>The person or persons who transcribe the source into Plaine &amp; Easie.</dd>
+    </dl>
+</section>
 <section class="normative">
     <h2>The Plaine &amp; Easie Code</h2>
     <section>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
-        <div class="issue" data-number="26"></div>
         <p>
-            The symbol <code>x</code> indicates sharp keys and <code>b</code> flat keys. The symbol is followed by the
-            capital letters that indicate the altered notes.
+            The character <code>x</code> indicates sharp keys and <code>b</code> flat keys. These characters MUST be
+            followed by a list of <a href="#note-names">note names</a> that indicate the altered notes.
         </p>
-
+        <p>
+            The list of note names in a key signature SHOULD follow the <a>circle of fifths</a> ordering. Missing
```
lpugin, line 204:
```suggestion
            The list of note codes in a key signature SHOULD follow the <a>circle of fifths</a> ordering. Missing
```

v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
+            accidentals in this list SHOULD be supplied by the <a>transcriber</a>. Note names MUST NOT be repeated.
```
lpugin, line 205:
```suggestion
            accidentals in this list SHOULD be supplied by the <a>transcriber</a>. Note codes MUST NOT be repeated.
```

v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
+        </p>
+        <p>
+            A key signature containing a single <code>n</code> character MAY be supplied to indicate a natural key
+            signature. This character MUST NOT be followed by any note names.
```
lpugin, line 209:
```suggestion
            signature. This character MUST NOT be followed by any note codes.
```

v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
+        </p>
+        <p>
+            A key signature MAY contain note names within square brackets, <code>[]</code>, to indicate that the
```
lpugin, line 212:
```suggestion
            A key signature MAY contain note codes within square brackets, <code>[]</code>, to indicate that the
```

v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
+            note names are not in the original source and have been supplied by the <a>transcriber</a>. Consecutively
```
lpugin, line 213:
```suggestion
            note codes are not in the original source and have been supplied by the <a>transcriber</a>. Consecutively
```

v2/index.html
```diff
@@ -180,21 +188,42 @@ <h3>Clef</h3>
...
+            supplied note names MUST be within a single set of brackets. A key signature MAY contain more than one
```
lpugin, line 214:
```suggestion
            supplied note codes MUST be within a single set of brackets. A key signature MAY contain more than one
```

lpugin (commented):
I am suggesting to use `note codes` instead of `note names`. But only as a suggestion.

ahankinson:
Since we would also need to change it in the actual "Note Names" section, outside of the scope of this change, I have opened #74 for this.

lpugin (approved):

PR #68, opened 2024-05-15
Fixed: Octave prose

### LINKED ISSUE #40: Improve octave values prose
The prose of the octave values section can be improved to set out the form and number of apostrophes and commas.

This should also be the place where it is specified that omitting the octave marker will keep the following notes in the same octave, and to switch octaves you need to explicitly give a new octave marker.

ahankinson:
Also fixed an issue that was preventing the examples from showing with Verovio.

Fixes #40

### DIFF
--- a/static/js/verovio-rendering.js
+++ b/static/js/verovio-rendering.js
@@ -1,5 +1,5 @@
 document.addEventListener("DOMContentLoaded", (event) => {
-    Module.onRuntimeInitialized = async _ => {
+    verovio.module.onRuntimeInitialized = async _ => {
         let tk = new verovio.toolkit();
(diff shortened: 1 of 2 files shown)


lpugin:
Looks good. The only question I have is whereas we expect at least one octave indication to be given in the incipit, or if a default (`'` ?) should be assumed.

ahankinson:
I think a default should be assumed, like we assume `4` for durations. I will update the prose to state that.

lpugin (approved):

ahankinson:
In the interests of moving things along, I've reduced the number of approvals required from 2 to 1...

PR #76, opened 2024-05-17
Fixed: Re-ordering and other structural changes

### LINKED ISSUE #69: Add a reference for Scientific Pitch Notation
Octave values are given according to Scientific Pitch Notation. 

https://en.wikipedia.org/wiki/Scientific_pitch_notation

We should add an appropriate reference and definition to the spec to clarify this.

ahankinson:
This PR is a bit of a monolith. It contains the following changes:

1. It restructures the order of sections. It did not make sense that grace notes came before beaming in the order. I understand that the ordering was (roughly) given in "encoding" order, but in practice this meant that fairly important sections, like chords, were placed last as they did not really fit.
2. It moves some of the "minor" encoding features to the "Other" section.
3. Some small tweaking of the octave sections, and an inclusion of a reference to scientific pitch

Fixes #69

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -46,11 +46,12 @@
             shortName: "paec",
             maxTocLevel: 4,
             localBiblio: {
-                "SCIENTIFIC-PITCH": {
-                    "authors": "Young, Robert W.",
-                    "title": "Terminology for Logarithmic Frequency Units",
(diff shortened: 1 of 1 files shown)


ahankinson:
The preview for the changes is here: https://plaine-and-easie.info/pr-preview/pr-76/v2/

lpugin (approved):

PR #79, opened 2024-05-21
Clarify that note names must be uppercase

ahankinson:
ASCII decimal codes are given for clarity, but not entirely sure they are needed. Feel free to suggest revision.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -528,8 +528,10 @@
                 A note name MUST be provided to indicate the pitch class of the encoded note.
             </p>
             <p>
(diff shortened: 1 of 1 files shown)


lpugin (commented):
I find giving the ASCII code in () confusing more than helpful. Saying that they should be ASCII characters seems enough to me.

ahankinson:
It’s more that they should be those exact characters (uppercase A) so that there is no ambiguity about using lower-case for note names.

ahankinson:
I've removed the ASCII codes

lpugin (approved):

PR #80, opened 2024-05-22
Fixed: Introduction to the Musical Notation section

### LINKED ISSUE #43: Add prose introduction to the "Music notation" section
Currently the "music notation" section launches straight into the octave symbols description, but it would be good to give an overview of the expected form of notation within the "body" of the PAE encoding.

Perhaps we can re-employ and update the visual structure provided by Massimo earlier (attached). 

pae-code

### LINKED ISSUE #56: Clarification of the order requirement for octave, duration, accidental and pitch
The specifications are currently not clear about a possible requirement regarding the order of these, except maybe through the [table](https://github.com/rism-digital/pae-code-spec/issues/43) of signs.

We need to decide if we want to restrict the code validity to a specific order, and if yes which one. The one given in the table looks logical. That is:
* octave
* duration
* accidental
* pitch

ahankinson:
- Defines the "Logical Unit" as a grouping of characters within the notation section
- Defines the required ordering of characters with a note "unit"

Fixes #43
Fixes #56

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -160,6 +160,9 @@
         <dd><!-- TODO --></dd>
         <dt><dfn>Transcriber</dfn></dt>
         <dd>The person or persons who transcribe the source into Plaine &amp; Easie.</dd>
+        <dt><dfn>Logical Unit</dfn></dt>
+        <dd>A logical unit of notation data consists of one or more characters that represent a single musical notation
+        figure.</dd>
     </dl>
 </section>
 <section class="normative">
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -107,6 +107,9 @@ <h2>Terminology</h2>
         <dd><!-- TODO --></dd>
         <dt><dfn>Transcriber</dfn></dt>
         <dd>The person or persons who transcribe the source into Plaine &amp; Easie.</dd>
+        <dt><dfn>Logical Unit</dfn></dt>
+        <dd>A logical unit of notation data consists of one or more characters that represent a single musical notation
+        figure.</dd>
```
lpugin, line 112: What is a musical notation figure? This is not clear, or am I lacking context?

v2/index.html
```diff
@@ -522,10 +525,51 @@ <h3>Musical Notation</h3>
...
+                A musical note is the single most complex logical unit within the <abbr>PAEC</abbr>. It consists of
+                one or more characters representing a note on a musical staff. Many characters representing a note
+                are optional, but where one or more character for a note occurs, they MUST occur in the following
+                order:
+            </p>
+            <figure>
+                <ol>
+                    <li>(optional) The grace note character, or the grace note group characters;</li>
+                    <li>(optional) The octave character(s);</li>
+                    <li>(optional) The duration character, followed by a duration dot character;</li>
```
lpugin, line 545: I think we need to keep the order between octave an duration flexible, but maybe this is for a later fix?

ahankinson: Do you have a use case where keeping the order flexible is useful?

lpugin: No. It is just than when typing it requires additional thinking to have a fix order. I would understand people would complain about it since. Also, there is no strong argument for having the octave before the duration - one might argue that octave relates to the pitch, so that duration should come first.

lpugin: I agree that we need to simplify data processing as much as possible. That would be the only exception for me from my (little) experience in cataloguing incipit. Maybe @jenniferward has a opinion on this?

jenniferward: I have been paying attention to the order of octave vs. duration for a while now and I do find that sometimes my brain says "OK, first octave" and sometimes it says "OK, first duration".  I think it happens when there is a significant switch: a big octave change or a big rhythmic change or if I have to coordinate one of them with other elements (curly brackets etc.). But I haven't kept track of when this happens, it's just my feeling.

Can we take a look at the data and find out how often `'''4` vs. `4'''` (or similar) happens? I think we can get used to a consistent order if prompted but I'd be curious what the natural tendency is.

lpugin: I think this is the main point. It really depends on the content of the music. Sometimes it clearly makes more sense to have the octave first, and sometimes the other way around. Running an analysis to see if there is a clear tendency towards octave-first would be good.

jenniferward: I meant that my fingers go naturally to either the octave first or the duration first, depending on musical context.

ahankinson:
I was talking this over with Catherine tonight, and she made a suggestion that makes sense, but comes a bit out of left field: we adopt a modified version of Helmholtz notation for pitch and octave.

https://en.m.wikipedia.org/wiki/Helmholtz_pitch_notation

That is, we notate it as duration-pitch-octave.

For the fourth octave and above this will map almost exactly to the standard Helmholtz notation, varying only by the case of the letters (we use uppercase).

For the lower octaves we keep the comma notation. Helmholtz indicates this by case or by a single comma in the contra octave, but I don’t think we can do this.

I like this suggestion because it aligns us with an existing standard for the ordering of note elements. But I can also see that it’s a bigger change than intended.

lpugin:
I think if we were inventing PAE that would be a good option. But yes, this I think would be too much of a change.

ahankinson:
I've created #97 and #98 to follow up on this PR, so (hopefully) it can be merged and we can address those issues later.

lpugin (approved):

PR #83, opened 2024-05-22
Fixed: Rests

### LINKED ISSUE #29: Clarify multi-measure rests must be followed by a barline
The spec currently says this about multi-measure rests:

> measure rest (followed by number of measures and a bar line)

This should be made more explicit and state that a multi-measure rest MUST be followed by a bar line.

### LINKED ISSUE #41: Provide form of rests
The rests section does not actually spell out the form of a rest with a duration. This should be added.

ahankinson:
- Clarifies forms of note rests and measure rests
- States that measure rests must be followed by a bar line

Fixes #29
Fixes #41

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -967,7 +967,9 @@
             <div class="issue" data-number="41"></div>
             <p>
                 Rests for single notes MUST be indicated by a hyphen/minus character <code>-</code>. This character
-                MUST be preceded by a <a>duration</a> value giving the musical duration of the rest.
+                MAY be preceded by a <a>duration</a> value giving the musical duration of the rest. If the duration
+                is omitted, the last specified duration is used. If no duration is supplied in the <a>encoding</a>
+                a default duration of <code>4</code> is assumed.
             </p>
             <p>
                 Measure rests MUST be indicated by an equal sign character <code>=</code>. This character MUST be
--- a/v2/index.html
+++ b/v2/index.html
@@ -965,27 +965,27 @@
             <h4>Rests</h4>
             <div class="issue" data-number="29"></div>
             <div class="issue" data-number="41"></div>
-            <p>Rests for single notes are indicated by <code>-</code> (a minus sign).</p>
-            <table class="simple">
-                <thead>
-                <tr>
-                    <th>Code</th>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -965,27 +965,27 @@ <h4>Accidentals</h4>
...
-                    <td><code>=</code></td>
-                    <td>measure rest (followed by number of measures and a bar line)</td>
-                </tr>
-
-                </tbody>
-            </table>
-
+            <p>
+                Rests for single notes MUST be indicated by a hyphen/minus character <code>-</code>. This character
+                MUST be preceded by a <a>duration</a> value giving the musical duration of the rest.
```
lpugin, line 970: This is not really true. A rest will just use the current duration if not duration is indicated

v2/index.html
```diff
@@ -965,27 +965,27 @@ <h4>Accidentals</h4>
...
+                Measure rests MUST be indicated by an equal sign character <code>=</code>. This character MUST be
+                followed by a non-negative integer indicating the number of measures for which this rest applies,
+                unless the measure rest only applies to a single measure. In this case, the number MAY be omitted.
+            </p>
+            <p>
+                Measure rests MUST be followed by a <a>bar line</a> character.
+            </p>
+            <p>
+                Measure rests MUST NOT be used with Mensural notation, due to the general absence of measures in
+                this notation.
```
lpugin, line 982: Fine for now, but might need to revise this depending on some clarifications for the mensural notation.

lpugin (approved):

v2/index.html
ahankinson: OK, I can change this.

v2/index.html
ahankinson: Which clarifications?

lpugin: What do we want to do with mensural notation in the RISM dataset. This would make some incipits in RISM invalid, so this is a problem.

ahankinson: It would make them invalid to version 2 of the spec, but not version 1. We can either keep those at V1, or upgrade them to V2 and add a bar line.

PR #91, opened 2024-05-23
Fixed: Specification of clef, keysig, timesig changes

ahankinson:


ahankinson:
Comment about this change from @lpugin on the errant commit of this section on another branch: 

> This [not permitting a space within the c/t/k change section] is not what Verovio currently expects. I think it would be clearer to require a space after every individual change.

I thought about that. My reasoning is this: We do not permit a space in any other part of the "Music notation" section. The clef / timesig / keysig (c/t/k) section is the only place where we permit spaces. 

There are a few factors I considered:

1. Most of the time, only one of the c/t/k changes will be required. So the issue of having multiple ones together is small.
2. When more than one change is required, the field delimiters are unique enough that you can use them as separators
3. A lot of the characters in the 'change' section can mean something different in the musical section. 

So the idea with a single space is to create a sort-of "header" or "scoreDef" section that is clearly separate from the "music notation" section. This is already the case with the "real" c/t/k sections, that they are set apart from the notation. If we allow (or even require) spaces between each of the "changes" too, it will be harder to tell what part is the change, and what part is the notation. 

And, for what it's worth, Verovio already handles this correctly even though it flags it with a warning.

lpugin:
OK. Having no space would make it consistent with the single-line input, so I think you are right. We need a ticket or label to remember we need to fix Verovio.

ahankinson:
Done https://github.com/rism-digital/verovio/issues/3692

lpugin (approved):

PR #95, opened 2024-05-26
Fixed: Write repeat group section

ahankinson:
Prose for the repeat group section

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1810,28 +1810,21 @@
         <section>
             <h4>Shortcuts</h4>
             <section>
-                <h5>Repeated figures</h5>
-                <table class="simple">
-                    <thead>
-                    <tr>
-                        <th>Code</th>
-                        <th>Name</th>
-                    </tr>
-                    </thead>
-                    <tbody>
-                    <tr>
-                        <td><code>!</code></td>
-                        <td>beginning and end of passage</td>
-                    </tr>
-                    <tr>
-                        <td><code>f</code></td>
-                        <td>repetition indication of the notes that appear within !...!</td>
-                    </tr>
-                    </tbody>
-                </table>
+                <h5>Repeat group</h5>
                 <p>
-                    The group will be repeated as many times as the <code>f</code> appears after the second
-                    <code>!</code>. Repetition is possible only within the same measure.
+                    If one or notes or rests are repeated several times, a repeat group
+                    MAY be used.
+                </p>
+                <p>
+                    A repeat group is only valid within a single measure. To repeat entire
+                    measures, use the <a>Measure repeat</a> shortcut.
+                <p>
+                <p>
+                    To use a repeat group, mark the beginning and the end of the group with
+                    an exclamation mark character <code>!</code>. The lower-case character
+                    <code>f</code> MUST immediately follow the ending <code>!</code>, and
+                    MUST occur the number of times the figure should be repeated. The
+                    <code>f</code> MUST be specified at least once.
                 </p>
                 <aside class="example" title="Encoding Repeated Figures">
                     <table class="simple" style="width: 100%">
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1623,28 +1623,21 @@ <h5>Tremolo, Slash</h5>
...
-                    </tbody>
-                </table>
+                <h5>Repeat group</h5>
+                <p>
+                    If one or notes or rests are repeated several times, a repeat group
+                    MAY be used.
+                </p>
+                <p>
+                    A repeat group is only valid within a single measure. To repeat entire
+                    measures, use the <a>Measure repeat</a> shortcut.
```
lpugin, line 1633: I don't think we need this restriction. This
```
!{'6CDEF}!f/ff/ff//
```
should (and does) work
image

ahankinson: That restriction was from the previous guidelines:

> Repetition is possible only within the same measure.

> `!{'6CDEF}!f/ff/ff//`

Isn't that the same as:

> `!{'6CDEF}!f/i/i//`

Do we want to encourage both?

ahankinson: Edit: That does work. I would say that if we want repeated measures, we should use measure repeats and not encourage repeat groups over measures.

lpugin: This is not the same. This is a quite typical use of group and measure repeat. 
```
!{'3CGEG}!f!{'DGFG}!f/i
```
image

Since this is not ambiguous, I don't see a problem with it

ahankinson: But that's two repeat groups within the same measure? It's not saying that there must only be one repeat group in a measure, it's saying that you can't pile `f`s on with a bar line in between them.

lpugin: What about this:
```
!{'6CDEF}!f/f4G/i/
```
image

Looks fine to me, so I am not sure what you are trying to forbid.

ahankinson: Yes, that is one instance where the repeat group is a slightly shorter way of writing it. I'm not sure it's clearer, though, since the `i` is repeating only one repetion in measure 1, and then only the second half of measure 2. 

> I am not sure what you are trying to forbid.

I'm trying to keep the number of rules and exceptions to those rules low so that they can be easily followed and interpreted. It's a lot easier to say "nothing should come between the `f`s, and the `f`s cannot cross the bar line" than to say "nothing should come between the `f`s except for bar lines, and you can repeat the `f`s as many times as you like except the `f` is only the repeat and not the full figure."

lpugin: It is shorter in this example because it is only a few note longs, but the patterns to repeat might be much longer. 

I think we only need to change ` A repeat group is only valid within a single measure.` to ` A repeat group remains valid in the following measure if it follows the barline immediately.`

ahankinson: Two last attempts to try and convince you:

1.) It's pretty clearly stated in Version 1 that the repeat group only goes to the end of the measure. So Verovio is technically wrong here. "Repetition is possible only within the same measure."
2.) In the original paper it isn't spelled out in full, but given the example I think it was intended that it only goes to the end of the measure. 

image

lpugin: OK. I will fix Verovio and @carlolic will fix the data. Please open and issue on both repositories.

It is still not clear to say `A repeat group is only valid within a single measure. To repeat entire measures, use the Measure repeat shortcut.` since the two repetition scopes are different. I don't understand why the two are bound together here and it would help not to mixed them up in the explanation.

ahankinson: Issues opened on both!

v2/index.html
```diff
@@ -1810,28 +1810,21 @@ <h5>Tremolo, Slash</h5>
...
```
lpugin, line 1820:
```suggestion
                    A repeat group is only valid within a single measure.
```
The repetition of an entire measure has nothing to do with repeat groups and it makes it very confusing to say this here.

lpugin (approved):

PR #99, opened 2024-05-27
Fixed: Prose for encoding grace notes

### LINKED ISSUE #22: Clarification with order of elements in ornaments
P&E says:

5. Grace notes and ornaments
g = acciaccatura (without rhythmic value, **precedes the note name**)
q = appoggiatura (with rhythmic value, **precedes the note name**)

but Verovio is not as strict: both ,8qB and ,q8B seem to be allowed.
Could you please clarify the order of elements here? Rhythm-q or should it be q-rhythm?

### LINKED ISSUE #75: Clarify whether `q` and `qq...r` can be used interchangeably, irrespective of the number of grace notes involved
The old specifications suggest that a single grace note should be preceded by a single `q`, while a group of grace notes belonging together should be preceded by `qq` and closed with `r`. That said, a number of incipits in Muscat ignore this principle and encode a  group by adding a `q` before each of its notes, and there are also instances of the opposite, namely `qq...r` groups with a single note in the middle (though these appear to be much less numerous and could thus have been simple oversights on the cataloger's part).

It should be decided whether we consider this cross-breeding as permissible (Verovio accepts both variants), and in any case, the specifications should use be rephrased to make our stance clear.

ahankinson:
Fixes #75
Fixes #22

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1599,7 +1599,7 @@
                     MAY occur.
                 </p>
                 <p>
-                    For multiple consecutive appogiatura, the appogiatura group MAY be used.
+                    For multiple consecutive appogiatura, the appogiatura group SHOULD be used.
                     The lower-case characters <code>qq</code> MUST be given as the first
                     characters in the first note of the group, and the lower-case character
                     <code>r</code> MUST be the last character in the final note of the group.
--- a/v2/index.html
+++ b/v2/index.html
@@ -1578,6 +1578,37 @@
                 <h5>Grace Notes</h5>
                 <div class="issue" data-number="22"></div>
                 <div class="issue" data-number="75"></div>
+                <p>
+                    There are two types of grace note: Acciaccatura and appogiatura.
+                </p>
+                <p>
+                    For a single acciaccatura the lower-case character <code>g</code>
+                    MUST be used. This character MUST be specified in the order given above;
+                    that is, it MUST precede the note name, and MUST also precede all other
+                    optional attributes of the note. Consecutive single acciaccatura
+                    SHOULD NOT occur.
+                </p>
+                <p>
+                    There are two methods used to encode appogiatura.
+                </p>
+                <p>
+                    For a single appogiatura, the lower-case character <code>q</code>
+                    MUST be used. This character MUST be specified in the order given above;
+                    that is, it MUST precede the note name, and MUST also precede all other
+                    optional attributes of the note. Consecutive single appogiatura
+                    MAY occur.
+                </p>
+                <p>
+                    For multiple consecutive appogiatura, the appogiatura group MAY be used.
+                    The lower-case characters <code>qq</code> MUST be given as the first
+                    characters in the first note of the group, and the lower-case character
+                    <code>r</code> MUST be the last character in the final note of the group.
+                    There MUST be more than one note in a appogiatura group.
+                </p>
+                <p>
+                    The <code>r</code> character MUST follow the closing beam character
+                    that is part of the appogiatura group.
+                </p>
                 <aside class="example" title="Encoding Grace Notes">
                     <table class="simple" style="width: 100%">
                         <thead>
@@ -1593,39 +1624,39 @@
                                 <script type="application/json">
                                     {
                                         "clef": "G-2",
-                                        "data": "'2A''gC{''8D'8B}''4C"
+                                        "data": "'4Ag''C{''8D'8B}"
                                     }
                                 </script>
-                                <code>g</code>
+                                <code>'4Ag''C{''8D'8B}</code>
                             </td>
                             <td class="notation-result"></td>
-                            <td>acciaccatura (without rhythmic value, precedes the note name)</td>
+                            <td>Acciaccatura (without rhythmic value)</td>
                         </tr>
                         <tr class="notation-example">
                             <td class="notation-code">
                                 <script type="application/json">
                                     {
                                         "clef": "G-2",
-                                        "data": "'2A'gB{'8A'8G}'4A"
+                                        "data": "'4Aq8'B{'8A'8G}"
                                     }
                                 </script>
-                                <code>q</code>
+                                <code>'4Aq8'B{'8A'8G}</code>
                             </td>
                             <td class="notation-result"></td>
-                            <td>appoggiatura (with rhythmic value, precedes the note name)</td>
+                            <td>Appoggiatura (with rhythmic value)</td>
                         </tr>
                         <tr class="notation-example">
                             <td class="notation-code">
                                 <script type="application/json">
                                     {
                                         "clef": "G-2",
-                                        "data": "'2A''qq{'8B''8C}r{''8D'8B}''4C"
+                                        "data": "'4Aqq''{'8B''8C}r{''8D'8B}"
                                     }
                                 </script>
-                                <code>qq..r</code>
+                                <code>'4Aqq''{'8B''8C}r{''8D'8B}</code>
                             </td>
                             <td class="notation-result"></td>
-                            <td>double appoggiatura, slide, or multiple appoggiatura (with rhythmic value)</td>
+                            <td>Double appoggiatura, slide, or multiple appoggiatura</td>
                         </tr>
                         </tbody>
                     </table>


v2/index.html
```diff
@@ -1520,6 +1520,37 @@ <h5>Tied Notes</h5>
...
+                </p>
+                <p>
+                    For a single appogiatura, the lower-case character <code>q</code>
+                    MUST be used. This character MUST be specified in the order given above;
+                    that is, it MUST precede the note name, and MUST also precede all other
+                    optional attributes of the note. Consecutive single appogiatura
+                    MAY occur.
+                </p>
+                <p>
+                    For multiple consecutive appogiatura, the appogiatura group MAY be used.
```
lpugin, line 1544:
```suggestion
                    For multiple consecutive appogiatura, the appogiatura group SHOULD be used.
```

ahankinson:
Good point, although I'm really not sure *how* they are interpreted? If I had a quarter note as a appogiatura, does that really take a full beat?

lpugin:
They are just visual durations, but you need them. Otherwise the current duration will apply. It is the opposite with acciaccature. `8` is always implied and no duration must be given (including 8 as it is now in Verovio, but we could relax this and not yield a warning for `4Cg8F4E`

v2/index.html
```diff
@@ -1578,6 +1578,37 @@ <h5>Tied Notes</h5>
...
+                </p>
+                <p>
+                    There are two methods used to encode appogiatura.
+                </p>
+                <p>
+                    For a single appogiatura, the lower-case character <code>q</code>
+                    MUST be used. This character MUST be specified in the order given above;
+                    that is, it MUST precede the note name, and MUST also precede all other
+                    optional attributes of the note. Consecutive single appogiatura
+                    MAY occur.
```
lpugin, line 1599:
```suggestion
                    optional attributes of the note. Consecutive single appogiatura
                    MAY occur even though they SHOULD be coded with an appogiatura group.
```

ahankinson: Updated it -- "SHOULD" implied "MAY", so the first part was redundant.

lpugin: It might be technically the same but being explicit draws the attention on the existence of appogiatura groups. I think it would be important to have a sentence recommending to use `qq...r` instead of `q.q.q.`.

ahankinson: Maybe something for the encoding guidelines?

lpugin: Not sure. The point is that there are not the same and that you can express more things with a `qq...r`. I would rather remove the sentence, or even forbid them and fix our data.

ahankinson: What's the difference between these two except that one is using single notes, and the other not? 

`A{q8Bq8''C}{'AB}/4A{qq8B''Cr}{'AB}`

image

I thought I would compare the MEI, but that only raised more questions than answers.
 
<details>

<summary>MEI Output</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title />
         </titleStmt>
         <pubStmt>
            <date isodate="2024-05-27-13:54:26" />
         </pubStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application xml:id="verovio" version="4.2.0-8a772f5">
               <name>Verovio (4.2.0-8a772f5)</name>
            </application>
         </appInfo>
         <projectDesc>
            <p>MEI encoded with Verovio</p>
            <p>Converted from Plaine and Easie to MEI</p>
         </projectDesc>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv xml:id="m15wwqth">
            <score xml:id="sv3lqdm">
               <scoreDef xml:id="s2gqwr3">
                  <staffGrp xml:id="s8j2ung">
                     <staffDef xml:id="s1ahluu1" n="1" lines="5" clef.shape="G" clef.line="2" />
                  </staffGrp>
               </scoreDef>
               <section xml:id="s1pfogn4">
                  <measure xml:id="mfnjo7e" right="single">
                     <staff xml:id="s1mj6nxn" n="1">
                        <layer xml:id="l1d0qcvg" n="1">
                           <note xml:id="nmwwo00" dur="4" oct="4" pname="a" />
                           <beam xml:id="b4ks4fv">
                              <note xml:id="n1s4ffnc" dur="8" oct="4" pname="b" grace="acc" stem.dir="up" />
                              <note xml:id="n9ixwzy" dur="8" oct="5" pname="c" grace="acc" stem.dir="up" />
                           </beam>
                           <beam xml:id="b147azdb">
                              <note xml:id="n1ffvwuh" dur="8" oct="4" pname="a" />
                              <note xml:id="nzosgrv" dur="8" oct="4" pname="b" />
                           </beam>
                        </layer>
                     </staff>
                  </measure>
                  <measure xml:id="mhbsl03" right="invis">
                     <staff xml:id="seeh8fz" n="1">
                        <layer xml:id="l5odh2d" n="1">
                           <note xml:id="n1q84ocf" dur="4" oct="4" pname="a" />
                           <beam xml:id="b1kzq82m">
                              <graceGrp xml:id="ga734np">
                                 <note xml:id="n1gzuhvg" dur="8" oct="4" pname="b" />
                                 <note xml:id="n1vw1hrc" dur="8" oct="5" pname="c" />
                              </graceGrp>
                           </beam>
                           <beam xml:id="bshell2">
                              <note xml:id="n36inxr" dur="8" oct="4" pname="a" />
                              <note xml:id="na56g6u" dur="8" oct="4" pname="b" />
                           </beam>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

</details>

 - There is no indication in the `<graceGrp>` of what kind of grace note it is. I'm guessing `<graceGroup>` should also provide for `@grace`?
 - Or do the notes inside of `<graceGrp>` need to be marked with `@grace`? How do you know the `<graceGroup>` is an appogiatura?
 - `@ornam` is also available? When would you use `@ornam="A"` vs `@grace="acc"`? (Also it's confusing that "Appogiatura" is `acc`, since you might otherwise think that it is "acciaccatura" unless you know that it's actually "Accented")

Still, it's not clear to me how they are different, other than they are encoded differently, and that one should be preferred over the other.

lpugin (approved):

BaMikusi:
Sorry to be behind with all of this, but there was no way for me to even look at this flood of GitHub notifications arriving in the past few days... In any case, IMO the SHOULD is too forgiving in both of these sentences:
(1) Consecutive single acciaccatura SHOULD NOT occur.
(2) Consecutive single appogiatura SHOULD NOT occur.

With (1), I don't even see how a series of consecutive `g` acciaccature could be visually rendered in a meaningful way.

And if with (2) we do not complete close the gate before those willing to use the grace notes for the rendering of entire incipits in lower case (so to speak), we shall keep seeing in our incipits diverse cues erroneuosly integrated from the other parts.

So, I would argue, we should simply say MUST in both cases.

BaMikusi:
... and then, of course, in the related sentences should also be adjusted:
- Multple appogiatura notes SHOULD be encoded using an Appogiatura Group
- For multiple consecutive appogiatura notes, the appogiatura group SHOULD be used.
I.e. (IMHO) in both cases rather MUST in place of SHOULD.

(BTW, at the beginning of the first of the above sentences there is a typo in "Multple".)

ahankinson:
Consecutive Acciaccature were disallowed here:

https://github.com/rism-digital/pae-code-spec/pull/126/files

 It's still open, though, so the change hasn't gone through yet.

ahankinson:
>  we do not complete close the gate before those willing to use the grace notes for the rendering of entire incipits in lower case (so to speak), we shall keep seeing in our incipits diverse cues erroneuosly integrated from the other parts.

Disallowing consecutive single appogiatura won't really stop this, since those could still be all encoded in consecutive groups. 

`qq44GGr/qq''DDr/qqEEr/qq2Dr/qq4CCr/qq'BBr/qqAAr/qq2Gr/`

BaMikusi:
You are right that this encoding would still be a way around, but I don't really see what we gain by allowing several single-note `q` characters one after the other, and so I think we should simply exclude this option -- and thereby still make the leak, through which such nonsense may creep in, a bit narrower.

PR #101, opened 2024-05-27
Fixed: First pass at neume notation

### LINKED ISSUE #4: Encoding Neumatic notation
**Summary**: As a software developer, I would like to see neumatic notation encoded differently, to make it easier to parse.

**Description**:

Currently neume notation is indicated by using the special `7.` note duration value. This means that parsers need to be prepared to encounter this value in any stream of notes, which makes parsing the note values more difficult even though encountering neume notation in the middle of a common notation note stream is highly unlikely.

A possible solution is to indicate neume notation by using the second character of a clef indicator, similar to mensural notation. For example, the colon `:` could be used, e.g., `G:2`. Other characters could be used as well. 

A preliminary check of the existing RISM incipits indicates that we currently have no incipits encoded this way, so the impact of this change will be small on this dataset. We should inquire more widely to assess the broader impact.

ahankinson:
The : character in the clef is now used to indicate neume notation, replacing the old '7.'

There may be other places in the guidelines that need fixing based on this change, but before doing too much it would be good to get this in place.

In light of this change, I have also clarified the meaning of the clefs with respect to octave, since neume clefs do not necessarily imply the octave in the same way. That probably needs to be worked through a bit more, but I thought it was a good idea to put the current practice in writing.

Fixes #4

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -199,8 +199,9 @@
                 be one of the values <code>G</code>, <code>g</code> (octave G), <code>C</code>, or <code>F</code>.
             </p>
             <p>
-                The second character MUST be one of the characters <code>-</code> to indicate modern notation, or
-                <code>+</code> to indicate mensural notation. (See: [[MensuralReference]])
+                The second character MUST be one of the characters <code>-</code> to indicate modern notation,
+                <code>+</code> to indicate mensural notation, or <code>:</code> to indicate neume notation. Each
+                of these notations impose different interpretations on the music notation in the incipit.
             </p>
             <p>
                 The third character MUST be a numeric value in the range 1-5, and indicates the reference staff line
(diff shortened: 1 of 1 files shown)


lpugin (approved):
Good for now. Can you open an issue [here](https://github.com/rism-digital/muscat-incipit-corrections/) to check the RISM data? We need to make sure we only neume incipits with only `7.` and nothing else.

I am not sure about the duration comments. I would be strict an say the neume notation must have no duration.

ahankinson:
> I would be strict an say the neume notation must have no duration.

It seemed strange to me to remove such an integral part of the note encoding.

lpugin:
> It seemed strange to me to remove such an integral part of the note encoding. 

It would make it very clear what the encoding is about. If one think it is a problem to have no duration, then I would make `4` as the only possible duration

ahankinson:
OK

lpugin (approved):

PR #104, opened 2024-05-28
Revision of the consequtive appogiature

lpugin:
Looking again at the current description, it seems to me that recommending not to use consecutive single appogiature in favour of the appogiatura group it better.


ahankinson (approved):

ahankinson:
Fixed! Thanks.

Issue #97, opened 2024-05-27
Determine order between duration and octave


ahankinson:
From #80 

Should we keep order of octave and duration flexible, or should there be a fixed order? TBD

ahankinson:
Worth noting that in the original paper by Brook and Gould, they give the order as:

<img width="780" alt="image" src="https://github.com/rism-digital/pae-code-spec/assets/163183/ebbd733c-90fd-4e90-bb48-6dffd81abd94">

BaMikusi:
This doesn't resolve the core problem but -- to my mind -- the actual pitch is defined by (1) the octave, (2) the accidental and (3) the note name. Hence, breaking up this logically coherent group by inserting the duration (i.e., something not pitch-defining) between (1) and (2) is not something I would want to prescribe as the only correct way.

Issue #60, opened 2022-10-21
Specify how to encode tied chords


carlolic:
In the spec (V1) there is no explanation how to encode tied chords (the problematic is already mentioned in V2). Some users overcame this by applying own rules:
- adding a + after each note in the chord, e.g. C+^E+^G+C^E^G
- adding a single + between the chords, e.g. C^E^G+C^E^G
Whereas the second approach would only work if the two chords are identical, the first approach would give more flexibility if not all the notes in the chord are tied. However it would be to specify whether + or ^ should directly follow the note.

ahankinson:
This should be considered with #2 and #1 as well. 

```
<CEG>_<CEG>
```

or, to tie only certain notes in a chord,

```
<C_E_G><CEG>
```

lpugin:
> or, to tie only certain notes in a chord,
> ```<C_E_G><CEG>```

I would not allow this. Only ties between the whole chords.

ahankinson:
Based on #112 and #110, the same mechanism for tied notes can be applied to tied chords. 

`^CEG>_`

Would be two C Major chords tied together. Much like the notes, the `_` would stand for the previously-stated chord, so it could also be part of other groups, or have its duration altered.

`^2xCxExG>/4_`

Would be a half-note C# Major chord tied over a barline to a quarter-note C# major chord. Note that this *doesn't* allow for repeating the accidentals in the following chord, so that's one limitation. But in a tied note the pitches are the same, so repeating is mostly a "cautionary" indicator.

This, however, would be possible:

`^2CEG>8(6{_^DxFA>^EGB>})`

![image](https://github.com/rism-digital/pae-code-spec/assets/163183/e1315a32-126b-4a0e-b9cc-0b7b1e61d389)

(Except that the first and second chords are tied)

PR #106, opened 2024-05-28
Editorial pass

ahankinson:
A first editorial pass over the spec.

Not many significant normative changes.

 - Grace notes groups were removed from the list of note features since they are a grouping mechanism and they follow their own rules
 - Similarly, the chord character was removed, since it functions more like a tie, and the order of ties and chords are not really established. (See #60, #2)
 - Specify that for neume notation, time sigs and key sigs MUST be omitted. This is following the decision to forbid durations on neumes, and the general direction of removing unnecessary attributes for specific notation types.

The ordered list of attributes on the notes was kept in the same order until #97 is resolved.


index.html
```diff
@@ -13,24 +13,24 @@
         </div>
 
         <p>
-            The Plaine &amp; Easie Code is a library standard that enables
-            entering music incipits in modern or mensural notation.
+            The Plaine &amp; Easie Code is a music notation encoding system
+            designed for the capture of melodic incipits: Short fragments of a
+            musical work that serve to uniquely identify that work.
```
lpugin, line 18:
```suggestion
            designed for the capture of melodic incipits: Short fragments of
            music that help identifying the music work.
```

ahankinson, line 18:
```suggestion
            designed for the capture of melodic incipits: Short fragments of
            music notation that help to identify the musical work.
```

lpugin (approved):

PR #107, opened 2024-05-28
Fixed: Write bar lines section

### LINKED ISSUE #44: Add prose description to the bar/measure lines section
Should spell out the possible values of bar and measure lines.

ahankinson:
Fixes #44

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1450,6 +1450,99 @@
         <section>
             <h4>Bar Lines</h4>
             <div class="issue" data-number="44"></div>
+            <p>
+                Bar lines MUST be indicated using one of the code options given in [[[#barlines-spec]]].
+            </p>
+            <p>
+                Bar lines SHOULD be inserted at intervals that correspond to the current time signature.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1450,6 +1450,99 @@ <h4>Beaming</h4>
...
             <h4>Bar Lines</h4>
             <div class="issue" data-number="44"></div>
+            <p>
+                Bar lines MUST be indicated using one of the code options given in [[[#barlines-spec]]].
+            </p>
+            <p>
+                Bar lines SHOULD be inserted at intervals that correspond to the current time signature.
+            </p>
+            <p>
+                Every incipit SHOULD end with a double bar line <code>//</code>.
```
lpugin, line 1460:
```suggestion
                Every incipit MAY end with a double bar line <code>//</code>.
```

ahankinson: If we're not recommending it (I thought we had discussed ending incipits with a double bar line at one point) then we should just remove it entirely. They MAY end it with any bar line, so only specifying the double line is weird.

Issue #65, opened 2023-05-25
Specify how to have a fermata over an entire chord


BaMikusi:
Looking at what Verovio (as implemented in Muscat) does, it seems that only a single note of a given chord should be bracketed, namely the first (i.e. the highest) one. This approach seems fine to me, but should explicitly be described in the specifications, since e.g. putting each note in brackets could also be a logical (though rather 'unplaine & uneasie') solution.

Issue #2, opened 2021-09-21
Encoding chord


lpugin:
The current specifications of PAE support the encoding of chords with the `^` character. 

From the current specifications

> Enter chords from the highest to the lowest note, each one separated by `^`.
> 
> Example: `''2D^'A^xF`

There are several issues with this approach:
* Parsing chords is problematic because the start of a chord is given retrospectively when the second note is encountered
* there are no clear specifications for tied chords. Since chords have to be encoded from the highest note to the lowest, the tie between two chords will always be (except for unison chords) between two notes with a different pitch, which is problematic.
* chords can have conflicting note durations
* how to apply a trill or a fermata to chord is not clearly defined

The proposal would be to change the chords to a container markup. To preserve some graphical similarity with the current `^`, a proposal would be to use `<` and `>`. These characters are not used elsewhere in PAE.

Example `2<''D'AxF>`

Some rules would be:
* Duration encoding should be given before the `<` chord start delimiter
* Chord delimiters should contain only pitch related information (octave, accidental and pitch name) 
* Ties should follow `>` chord end delimiters

It would need to be decided if fermata and trills should be outside or inside chord delimiters

lpugin:
It seems to be between 2) or 3). I still believe we can have a good V2 to which we can upgrade all the RISM data without dropping (significant amount of) data.

lpugin:
The I would stay way from both `<` AND `>`. Using only `>` actually does not make it better than using both. The theory is that you do not have to escape it because `<subfield>^ABC></subfield>` is valid. However, libraries might to `<subfield>^ABC&gt;</subfield>`, so in the end we have not avoided the problem. This is what happens with LibXML in ruby, at least.

ahankinson:
I think it’s ok if a library encodes any characters , but having an unencoded `<` is going to kill lots of things.

BaMikusi:
One point we haven't addressed above is the 'direction' for the encoding.

The present specifications ask the cataloger to enter the notes "from the highest to the lowest" -- a point that always gave me a pause, for I would instincitively do this the other way round, from bottom to top. Not simply on acoustical grounds, but rather because that should be the natural way to play chords on string instruments (and in fact even on the keyboard, if an arpeggio is expected, that will normally go from bottom to top -- unless there is a very explicite remark to do the contrary).

[...]

PR #110, opened 2024-05-28
Fixed chords

ahankinson:
This is a first attempt at reaching some agreement on the encoding of chords.

Chords are enclosed in ^ and >. The duration of the chord is given prior to the opening ^ (similar to tuplets). No duration values are allowed within the chord. (i.e., chords with notes with differing durations are not allowed.)

~Similar to repeat groups and repeat measures, tied chords are repeated with the v character.~ This has evolved, see #60  This means that you don't have to spell out the notes of the chords, and only identical tied chords are supported, varying only by their duration.

A p may be used to indicate a fermata over the chord.

Fixes https://github.com/rism-digital/pae-code-spec/issues/65
Fixes https://github.com/rism-digital/pae-code-spec/issues/60
Fixes https://github.com/rism-digital/pae-code-spec/issues/2

Examples:

`2^CEG>`

image

`{8^CEG>^DxFA>}`

image

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1363,7 +1363,7 @@
             <p>
                A <a>duration</a> value MAY precede the <code>^</code>. Two or more <a>note names</a>
                (with optional octave and accidental indications) MUST immediately follow this character.
-               The notes MUST be ordered from the lowest to the highest on the staff line. A <code>></code>
+               The notes SHOULD be ordered from the lowest to the highest on the staff line. A <code>></code>
                MUST be used to end the chord. A duration value for the individual notes within a chord MUST
                NOT be supplied.
             </p>
--- a/v2/index.html
+++ b/v2/index.html
@@ -1382,7 +1382,7 @@
             </p>
             <p>
                 The <code>v</code> MUST follow the closing <code>></code>. It MUST not be separated from
-                the chord by any character EXCEPT a <a>bar line</a>. It MAY be repeated as many times as
+                the chord by any character EXCEPT <a>bar line</a> characters. It MAY be repeated as many times as
                 necessary, and each repetition MAY change its duration.
             </p>
             <aside class="example" title="Encoding Chords">
--- a/v2/index.html
+++ b/v2/index.html
@@ -1358,7 +1358,29 @@
             <div class="issue" data-number="60"></div>
             <div class="issue" data-number="65"></div>
             <p>
-                Enter chords from the highest to the lowest note, each one separated by <code>^</code>.
+                Notes in a chords are enclosed in the characters <code>^</code> and <code>></code>.
+            </p>
+            <p>
+               A <a>duration</a> value MAY precede the <code>^</code>. Two or more <a>note names</a>
+               (with optional octave and accidental indications) MUST immediately follow this character.
+               The notes MUST be ordered from the lowest to the highest on the staff line. A <code>></code>
+               MUST be used to end the chord. A duration value for the individual notes within a chord MUST
+               NOT be supplied.
+            </p>
+            <p>
+                If the duration is omitted, the last specified duration is used. If no duration is
+                supplied in the <a>encoding</a> a default duration of <code>4</code> is assumed.
+            </p>
+            <p>
+                A <code>v</code> MAY be used to indicate a tied chord. A duration value MAY immediately
+                precede the <code>v</code>. If the duration is omitted, the last specified duration is
+                used. If no duration is supplied in the <a>encoding</a> a default duration of <code>4</code>
+                is assumed.
+            </p>
+            <p>
+                The <code>v</code> MUST follow the closing <code>></code>. It MUST not be separated from
+                the chord by any character EXCEPT a <a>bar line</a>. It MAY be repeated as many times as
+                necessary, and each repetition MAY change its duration.
             </p>
             <aside class="example" title="Encoding Chords">
                 <table class="simple" style="width: 100%">
@@ -1377,7 +1399,19 @@
                                     "data": "''2D^'A^xF"
                                 }
                             </script>
-                            <code>''2D^'A^xF</code>
+                            <code>2^'AxF>/2^AxF>/</code>
+                        </td>
+                        <td class="notation-result"></td>
+                    </tr>
+                    <tr class="notation-example">
+                        <td class="notation-code">
+                            <script type="application/json">
+                                {
+                                    "clef": "G-2",
+                                    "data": "''2D^'A^xF"
+                                }
+                            </script>
+                            <code>2^'AxF>/2v/4v//</code>
                         </td>
                         <td class="notation-result"></td>
                     </tr>
--- a/v2/index.html
+++ b/v2/index.html
@@ -1372,6 +1372,9 @@
                 supplied in the <a>encoding</a> a default duration of <code>4</code> is assumed.
             </p>
             <p>
+                A <code>p</code> character MAY follow the duration to indicate the chord has a fermata.
+            </p>
+            <p>
                 A <code>v</code> MAY be used to indicate a tied chord. A duration value MAY immediately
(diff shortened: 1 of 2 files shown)


v2/index.html
```diff
@@ -1358,14 +1358,40 @@ <h4>Chords</h4>
...
+                supplied in the <a>encoding</a> a default duration of <code>4</code> is assumed.
+            </p>
+            <p>
+                A <code>p</code> character MAY follow the duration to indicate the chord has a fermata.
+            </p>
+            <p>
+                A <code>v</code> MAY be used to indicate a tied chord. A duration value MAY immediately
+                precede the <code>v</code>. If the duration is omitted, the last specified duration is
+                used. If no duration is supplied in the <a>encoding</a> a default duration of <code>4</code>
+                is assumed.
```
lpugin, line 1381: Would this also require the `_`? That is `4^CEG>_v`? I would expect it because it makes it clearer.

v2/index.html
```diff
@@ -1358,14 +1358,40 @@ <h4>Chords</h4>
...
+               The notes MUST be ordered from the lowest to the highest on the staff line. A <code>></code>
+               MUST be used to end the chord. A duration value for the individual notes within a chord MUST
+               NOT be supplied.
+            </p>
+            <p>
+                If the duration is omitted, the last specified duration is used. If no duration is
+                supplied in the <a>encoding</a> a default duration of <code>4</code> is assumed.
+            </p>
+            <p>
+                A <code>p</code> character MAY follow the duration to indicate the chord has a fermata.
```
lpugin, line 1375: I guess we also want a trill? Not very likely, but I am sure somebody will find one

v2/index.html
```diff
@@ -1358,14 +1358,40 @@ <h4>Chords</h4>
...
+            <p>
+                A <code>v</code> MAY be used to indicate a tied chord. A duration value MAY immediately
+                precede the <code>v</code>. If the duration is omitted, the last specified duration is
+                used. If no duration is supplied in the <a>encoding</a> a default duration of <code>4</code>
+                is assumed.
+            </p>
+            <p>
+                The <code>v</code> MUST follow the closing <code>></code>. It MUST not be separated from
+                the chord by any character EXCEPT a <a>bar line</a>. It MAY be repeated as many times as
+                necessary, and each repetition MAY change its duration.
```
lpugin, line 1386: What about a chord tied to a tuplet chord? or a beam chord?

v2/index.html
ahankinson: No, the `_` is not necessary in this scheme. I was toying with the idea of using it, but it seemed like it might cause problems because it wouldn't function the same as ties, as they are currently specified, do. 

I want, effectively, a "repeat the chord" character, so that there is no possibility that the chords on either end of the tie are different. That's what the `v` gives. 

For notes, the current method means you have to specify the note on the other end. `F_F`. If we wanted to apply this to the chord, then I would expect `^CEG>_^CEG>`, not just `^CEG>_v`. Also, the `v` might get confused as a repeat character for the chord... it's not, strictly speaking, since you can't simply do ^CEG>v` and get a repeated chord with no ties.

ahankinson: A new proposal is in #112 that would unify the behaviour of ties, and allow the use of `_` on chords.

v2/index.html
```diff
@@ -1358,14 +1358,40 @@ <h4>Chords</h4>
             <div class="issue" data-number="60"></div>
             <div class="issue" data-number="65"></div>
             <p>
-                Enter chords from the highest to the lowest note, each one separated by <code>^</code>.
+                Notes in a chords are enclosed in the characters <code>^</code> and <code>></code>.
```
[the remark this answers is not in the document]
ahankinson: I'm a big -1 on using the `<` character.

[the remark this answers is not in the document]
ahankinson: I don't think we use the ampersand, do we?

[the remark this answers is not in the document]
lpugin: You are right, we do not use it. We should then ;-)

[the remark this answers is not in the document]
ahankinson: It seems counter-productive to phase out `+` because of encoding issues, but introduce `<` and `&` which will only bring similar issues...

v2/index.html
```diff
@@ -1358,14 +1358,40 @@ <h4>Chords</h4>
...
+            </p>
+            <p>
+               A <a>duration</a> value MAY precede the <code>^</code>. Two or more <a>note names</a>
+               (with optional octave and accidental indications) MUST immediately follow this character.
+               The notes MUST be ordered from the lowest to the highest on the staff line. A <code>></code>
```
ahankinson, line 1366:
```suggestion
               The notes SHOULD be ordered from the lowest to the highest on the staff line. A <code>></code>
```

lpugin:
There is one more thing that needs clarification - I think - which is if an octave change within a chord remains as current after it. I would expect it, but I don't think this is explicitly stated. Is it? Or do we want to change the this? We would need to think about the implications if we do.

I noticed this to be particularly "nasty" with the use of repetition. For example, if you do:

`2'G4!G^B^''D!f/` you probably expect the last chord to be repeated like that:
image

However, this is what Verovio produces - keeping the last octave as current:
image

ahankinson:
It's not explicitly stated. Could you please make an issue?

BaMikusi:
Apologies to @ahankinson for mixing the topics, but I have really lost the overview, and would not like to potentially duplicate tickets. In any case, @lpugin's example reminds me of a more general problem with repetition shortcuts.

Both with repeated measures (`i`) and repeat groups (`!...!f`) the intention seems to be to indicate a repeat _of the music_, but in fact only the code will get repeated, which might result in non-identical renderings, depending on the context (surrounding indications of octaves and durations). I guess in the V2 specs we should make clear once and for all _what_ gets repeated in such a case.

(Would actually be nice if that could be the actual music, but I guess that should prove a considerably bigger challenge for the rendering.)

ahankinson:
Could you clarify how you're differentiating the "code" and the "music"?

BaMikusi:
I have a source before me. I am encoding it, and at some point I realize that what comes (say, an entire bar) is exactly what I have just transcribed. PAE spec then encourages me to simply put an `i` in the next bar, but what Verovio then renders may still not be the same music/notation as what I had in that previous bar. Because the `i` in fact implied the repetition of the code, which is likely not quite what the cataloger meant to be repeated, when seeing a musical repetition in the source itself.

BaMikusi:
So, the specs should actually say: you could try to repeat the previous bar by entering an `i`, and if the circumstances are umproblematic, that will indeed work out for you.

BaMikusi:
Sorry, when writing my last comment I was just rushing out of the office to catch my train, and had no time come up with an example -- but now here's one:

Say, this is the music I actually see in the source:
'EF/GB''2C/'4GB''2C/
grafik

However, while transcribing this, I realize that the 2nd full bar is actually identical to the first one, so I decide to follow the specs and use an `i` shortcut like this:
'EF/GB''2C/i/
But while this `i` indeed appears effectively to copy over the code from one bar to the next, since at the beginning of the 1st full bar we were in the `'` octave and had quarter durations 'inherited' from the precedents, while at the start of the 2nd full bar we are in the `''` octave and have had a half value before, the music displayed will be so different that no one would guess this was meant to be a note for note repetition of the previous bar:
grafik

ahankinson:
@lpugin I think this is ready to be merged, but could you please double-check?

v2/index.html
ahankinson: I've created #148 -- I'd like to get this PR merged before it gets too far out of sync with `main` but yes, a trill should probably be allowed.

lpugin (approved):
We need to check the tied chords, but we should do this after.

Issue #89, opened 2024-05-23
Fix Verovio code


ahankinson:
In the ties section of V2 the ties are illustrated with underscores (reflecting #1 ) but are given in the code section with `+` so that they display correctly. Once Verovio is updated to support underscores with ties, this should be fixed. 

Left here as a reminder to fix this.

ahankinson:
Also the fermata and appogiatura groups are notated differently than the Verovio code.

PR #116, opened 2024-05-29
Fixed: Specify bar for alternating time sigs

### LINKED ISSUE #103: Separation of alternating time signatures
Currently alternating time signatures are separated by a space. However, this may be a challenge when processing changes to time signatures, since the space character is used to separate changes from the following music notation. 

Another character should be used instead; I am proposing the use of the pipe character (`|`). This character is commonly used to indicate "OR" in regular expressions (e.g., a|b  == "a OR b"). 

I understand that the pipe character is not that easy to use on some keyboards. However, alternating time signatures are also not very common. As best as I can tell, there are roughly 40 incipits in RISM Online that feature an alternating time signature. (0.002% of all incipits)

ahankinson:
The space character was problematic as it caused ambiguities with change of clef.

This specifies that alternating time signatures should be separated by a vertical bar

Fixes #103

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -483,8 +483,6 @@
             <p>
                 There are two main categories of time signature forms, <a>Common Western Music Notation</a>
                 (<abbr title="Common Western Music Notation">CWMN</abbr>) and Mensural.
-                If a mensuration sign is specified, the <a>clef</a> MUST specify a <code>+</code> separator to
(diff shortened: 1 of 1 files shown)


lpugin:
Relates to https://github.com/rism-digital/pae-code-spec/issues/89

lpugin (approved):

PR #118, opened 2024-05-30
Overhaul structure

ahankinson:
This commit is the first pass at overhauling the structure to make it flow logically. A second commit after this one will adjust the prose to fit the new structure, but this commit is primarily moving things around.

The "Plaine and Easie Code" section was removed, and the two main sections within promoted up: Staff Definitions and Musical Notation.

Within Musical Notation, the sections were moved around. Now, things that are attributes of notes and rests are grouped under the "Notes and Rests" section; Grouping mechanisms like beams or tuplets are in their own section, and staff elements, like bar lines, measure rests, and changes, are in their own section.

There may still be a little bit of re-ordering in the notes and rests section to get a consistent flow.

This effort is in response to the need to figure out some higher-level concepts in the PAEC, such as what a "note" is, and what attributes are part of a "note".

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -191,722 +191,579 @@
         figure.</dd>
     </dl>
 </section>
-<section class="normative">
-    <h2>The Plaine &amp; Easie Code</h2>
+<section>
     <div class="issue" data-number="57"></div>
+    <h2>Staff Definitions</h2>
+    <p>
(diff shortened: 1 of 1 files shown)


ahankinson:
Current structure:

image

lpugin (approved):

PR #127, opened 2024-06-03
Fixed: Expand duration group prose

ahankinson:
Expands the rules for repeated rhythmic sequences. Nothing has changed except a clarification on how they should be interpreted.

Figuring out where they could and couldn't be used took a bit of time; I think I got it, but may have missed something.

I renamed them "Duration Groups" because they are initiated by a grouping of duration values.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -2125,11 +2125,11 @@
             <h5>Duration Group</h5>
             <p>
                 Multiple duration values MAY be declared together. These values form a Duration Group, and MAY be
-                used to indicate repeating rhythmic patterns for the notes that follow.
+                used to indicate repeating rhythmic patterns for a sequence of notes.
             </p>
             <p>
-                The number and order of duration values MUST apply to an equivalent number of notes that follow.
-                If the number of notes that follow exceeds the number of duration values given in the Duration
+                The number and order of duration values MUST apply to an equivalent number in the sequence of notes.
+                If the number of notes in the sequence exceeds the number of duration values given in the Duration
                 Group, the duration values MUST repeat from the first.
             </p>
             <p>
@@ -2137,32 +2137,33 @@
                 MAY be given at any point in the note sequence, and will cancel the duration group.
             </p>
             <p>
-                Alterations to pitch, octave, beaming, and trills, and fermata MAY occur on successive repeats. A rest
-                MAY also occur within the notes that follow, with the appropriate duration value applied.
+                Alterations to pitch, octave, beaming, and trills, and fermata MAY occur on successive repeats of the
+                note sequence. A rest MAY occur within the note sequence, with the appropriate duration value applied.
             </p>
             <p>
-                A tie MAY be indicated in the notes that follow the Duration Group declaration.
+                A tie MAY be indicated to join two or more notes in the note sequence. The duration of each member note
+                in the tie MUST follow the duration order in the Duration Group.
             </p>
             <p>
                 A measure rest MAY occur within the notes that follow. The notes following the measure rest MUST
-                continue the application of the Duration Group sequence from the order of the sequence prior to
-                the measure rest.
+                continue the application of the durations in sequence.
             </p>
             <p>
-                Changes to Clef, Key Signature, and Time Signature MAY also occur within the repeating duration group.
-                The notes following the change MUST continue the application of the durations in sequence.
+                Changes to Clef, Key Signature, and Time Signature MAY occur within the note sequence. The notes
+                following the change MUST continue the application of the durations in sequence.
             </p>
             <p>
                 A chord MAY be used within the notes that follow a Duration Group. All notes in the chord MUST have
                 a single value from the Duration Group.
             </p>
             <p>
-                A tuplet MUST NOT be used as a duration group declaration. Use of a tuplet MUST cancel the Duration
-                Group sequence.
+                A tuplet MUST NOT be used in a duration group declaration. Use of a tuplet in the note sequence MUST
+                cancel the Duration Group sequence.
             </p>
             <p>
-                Acciacciatura, appogiatura, and appogiatura groups MUST NOT occur within a Duration Group. Use of these
-                ornaments MUST cancel the Duration Group.
+                Appogiatura and appogiatura groups MUST NOT occur within a Duration Group. Use of these ornaments
+                MUST cancel the Duration Group. Acciaccatura MAY occur with the notes of the group, but the
+                durations in the Duration Group MUST NOT apply to any note marked as acciaccatura.
             </p>
             <aside class="example" title="Encoding Rhythmic Sequences">
                 <table class="simple" style="width: 100%">
--- a/v2/index.html
+++ b/v2/index.html
@@ -2122,10 +2122,47 @@
             </aside>
         </section>
         <section>
-            <h5>Rhythmic sequence</h5>
+            <h5>Duration Group</h5>
             <p>
-                When the same rhythmic sequence is repeated, the sequence of rhythmic values can be stated once
-                before the note names.
+                Multiple duration values MAY be declared together. These values form a Duration Group, and MAY be
+                used to indicate repeating rhythmic patterns for the notes that follow.
+            </p>
+            <p>
+                The number and order of duration values MUST apply to an equivalent number of notes that follow.
+                If the number of notes that follow exceeds the number of duration values given in the Duration
+                Group, the duration values MUST repeat from the first.
+            </p>
+            <p>
+                A duration group MUST remain in effect until another duration value is declared. A new duration value
+                MAY be given at any point in the note sequence, and will cancel the duration group.
+            </p>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -2122,10 +2122,47 @@ <h5>Repeated measures</h5>
...
             <p>
-                When the same rhythmic sequence is repeated, the sequence of rhythmic values can be stated once
-                before the note names.
+                Multiple duration values MAY be declared together. These values form a Duration Group, and MAY be
+                used to indicate repeating rhythmic patterns for the notes that follow.
+            </p>
+            <p>
+                The number and order of duration values MUST apply to an equivalent number of notes that follow.
+                If the number of notes that follow exceeds the number of duration values given in the Duration
+                Group, the duration values MUST repeat from the first.
```
lpugin, line 2133: I don't understand this sentence.

v2/index.html
```diff
@@ -2122,10 +2122,47 @@ <h5>Repeated measures</h5>
...
+            </p>
+            <p>
+                A duration group MUST remain in effect until another duration value is declared. A new duration value
```
lpugin, line 2136:
```suggestion
                A duration group remains in effect until another duration value is declared. A new duration value
```

v2/index.html
```diff
@@ -2122,10 +2122,47 @@ <h5>Repeated measures</h5>
...
+                MAY be given at any point in the note sequence, and will cancel the duration group.
+            </p>
+            <p>
+                Alterations to pitch, octave, beaming, and trills, and fermata MAY occur on successive repeats. A rest
+                MAY also occur within the notes that follow, with the appropriate duration value applied.
```
lpugin, line 2141: As it stands in Verovio `g` is allowed within group since it does not introduce a new duration. `q` or `qq` is not. I think we should keep this.

v2/index.html
```diff
@@ -2122,10 +2122,47 @@ <h5>Repeated measures</h5>
...
+                continue the application of the Duration Group sequence from the order of the sequence prior to
+                the measure rest.
+            </p>
+            <p>
+                Changes to Clef, Key Signature, and Time Signature MAY also occur within the repeating duration group.
+                The notes following the change MUST continue the application of the durations in sequence.
+            </p>
+            <p>
+                A chord MAY be used within the notes that follow a Duration Group. All notes in the chord MUST have
+                a single value from the Duration Group.
```
lpugin, line 2157: The second sentence is not very clear. Chords have a single value anyway, haven't they?

v2/index.html
```diff
@@ -2122,10 +2122,47 @@ <h5>Repeated measures</h5>
...
+            </p>
+            <p>
+                A tuplet MUST NOT be used as a duration group declaration. Use of a tuplet MUST cancel the Duration
+                Group sequence.
+            </p>
+            <p>
+                Acciacciatura, appogiatura, and appogiatura groups MUST NOT occur within a Duration Group. Use of these
+                ornaments MUST cancel the Duration Group.
```
lpugin, line 2165: See above. Acciaccatura are currently supported and are not problematic since they have no duration

v2/index.html
ahankinson: It just means that the duration cycle... `484ABCDEFGAB` would result in the three repetitions of the `484` figure.

v2/index.html
ahankinson: Why did you remove this requirement? Without it, there is no requirement that a duration value interrupts the repetition of the group.

v2/index.html
ahankinson: That's never explicitly stated.

v2/index.html
lpugin: I am confused with the use of `MUST` here. My understanding is that it is about how things must be coded, not how what is written must match the code, so the `MUST` is confusing in this paragraph.

v2/index.html
lpugin: Same as above. I find this confusing.

v2/index.html
lpugin: I was thinking about [this](ttps://github.com/rism-digital/pae-code-spec/pull/110)

v2/index.html
lpugin: It is like if you say "a note following a `4` MUST be a quarter note"

ahankinson: It could a) stop (i.e., not repeat the durations) or b) go backwards (weird, I know, but it *could*) or c) be interruptible by adding a specific duration (which is why the requirement to cancel it is necessary) or d) apply to only otherwise grouped notes (e.g., only beams would take the duration group). 

It's better to be pedantic than to be vague, IMO. There's no room for interpretation when is says it MUST repeat from the first.

BTW, we do say that a note following a 4 MUST be a quarter note... https://plaine-and-easie.info/v2/index.html#durations

v2/index.html
ahankinson: Could you suggest some other wording that keeps the normative requirement that a duration group remains in effect until cancelled?

lpugin: Something like `"A duration group indicates that the note names following it have a duration as given in the group and in that order. When the number of notes reaches the number of durations given in the group, the durations applies again from the beginning of the group. Any duration given in the notes following the duration group cancels it."

ahankinson: I think we still need the MUST in there to indicate the requirement.

BaMikusi:
This is just to express my full support for @lpugin's view that the MUST's are quite confusing.
But perhaps an alternative could be to say "will remain" and "will have"?

v2/index.html
lpugin: > BTW, we do say that a note following a 4 MUST be a quarter note...

Can you point me to the line? Sorry, I cannot find it.

ahankinson:
"will" is a statement of fact, not a statement of requirement. "I will go to the store to find milk" is different from "I must go to the store to find milk." It can be made a requirement with the imperative "shall"  as in "you shall go to the store and find milk". 

I've been using MUST because it imposes a very easy system of understanding what things are required (MUST / MUST NOT), recommendations (SHOULD / SHOULD NOT), and optional (MAY). Otherwise there is a danger that internally the requirements are not phrased in a consistent way. 

The number of RFC keywords in use can be expanded if you think it would make for better prose, but (at least for me) the words are used to help me keep track of what the actual valid / invalid requirements are, and limiting them helps me to keep a consistent tense, tone, and terminology.

See RFC 2119 for a definition of what each of the terms mean: https://datatracker.ietf.org/doc/html/rfc2119

(Otherwise we end up with the current state of affairs where no one actually knows what "valid" PAE is other than "does it render OK, it must be fine.")

lpugin:
Sure, I get that, and I am perfectly fine where we have it. It is only that here you use it the other way around. Namely to say that the notes MUST be in accordance with the code. 

You do not say that notes after a `G-2` clef MUST be in `G-2` clef until a clef change is given. Or we say "The clef code MUST be three characters long", but then "The character `x` indicates sharp keys and `b` flat keys.", and not "The note with a character `x` in front MUST be a sharped note".

BaMikusi:
To approach the point from another angle, if *human* users will not understand your language (and if Laurent and I both stop short here, there would surely be others), it won't really matter how consistent your text appears to be from a structural point of view.

ahankinson:
OK; could you help me rewrite it in another way that makes it clear that there are actually some requirements for a duration group? I understand not liking the way it's phrased, but I also need a way of ensuring that someone coming to this that wants to implement support for PAE also knows what their tool MUST or MUST NOT do.

lpugin:
I'll use this as a model https://plaine-and-easie.info/v2/#repeat-group

ahankinson:
All users of this spec **are** _human users_. It's actually being written specifically FOR human users, who have an annoying tendency to get "creative" with their interpretations when no creativity is desired. If we don't want to end up "cleaning up" the incipits every decade or so because of creative interpretations of the rules, then this is the best chance we have to do so. And the best way I know how to do that is to create a very clear indication of, specifically, what IS and what IS NOT allowed. Not everyone needs to know this, but the people (us, and the "us" in the future) who create the tools that put the red squiggles under the things that are wrong **do** need to know, and they'll need this document to tell them what everyone else is supposed to be doing too.

If you mean the actual cataloguers looking for guidelines on how to apply these rules, then we have #50 "Create encoding guidelines" which we will write once we settle on the rules. However, #50 also needs to point to something as the authority for what is and isn't allowed.

ahankinson:
Sorry -- it was simply the emphasized "for humans" that triggered me. I know it's not a particularly natural way of writing, and for those who don't deal with specifications it can even be a bit alien, but it's my core intention that it is clearly written for humans first and foremost.

BaMikusi:
Alles klar -- and then apologies for that _human_ from my side.
After all, we are all humans, and if a human writes 20+ emails a day, plus 30+ Slack messages, surely some points will not be made elaborately enough.

PR #130, opened 2024-06-04
Fixed: Expand ligatures section

### LINKED ISSUE #28: Use of `u` to encode ligatures
We have many incipits where `+` has been used for encoding ligatures instead of ties.
* For representing ligatures in mensural notation
* For representing ligatures with stem-less notes (`7.`)

We need to clarify if this is supported or not. If yes, tooling needs to be adjusted accordingly. We should also consider adjusting the specifications to use another sign.

(We also have cases where `+` has been used to represent slurs. This is simply wrong.)

ahankinson:
Not too sure about the "aside" here, but I think we should have something to say that, in this case, you're not actually coding the visual (that is, the thing that you see on the page).

It could be moved to the guidelines, if it's deemed superfluous here.

Fixes #28

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1354,8 +1354,68 @@
             <div class="issue" data-number="28"></div>
             <h4>Ligatures</h4>
             <p>
-                Todo.
+                The presence of two notes joined in a ligature may be indicated with the lower-case character
+                <code>u</code>.
             </p>
+            <p>
+                Ligatures MAY occur between two notes. Several consecutive notes MAY be joined in a single ligature by
+                adding the <code>u</code> between each note.
+            </p>
+            <p>
+                A ligature MUST NOT occur between a note and a rest, and MUST NOT occur between a note and a group
+                of notes. (i.e., a beam, tuplet, or chord.) Ligatures also MUST NOT occur within a beam, tuplet,
+                or chord.
+            </p>
+            <p>
+                Ligatures MAY be used to indicate two notes are part of the same Neume.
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1354,8 +1354,68 @@ <h4>Ties</h4>
...
+                                {
+                                    "clef": "G-2",
+                                    "data": ""
+                                }
+                            </script>
+                            <code>8{ABqq{''CD}r'AB}</code>
+                        </td>
+                        <td class="notation-result"></td>
+                        <td>Beaming with a Grace Note Group</td>
+                    </tr>
```
lpugin, line 1415: Why is this example here?

v2/index.html
```diff
@@ -1354,8 +1354,68 @@ <h4>Ties</h4>
...
+            </p>
+            <aside class="note">
+                <p>
+                    While some notation renderers may attempt to reproduce the shape of the ligature in its original
+                    form, it is also common practice to indicate that two notes are in a ligature by including them
+                    separately but within a horizontal bracket. Both forms are valid visualizations of ligatures,
+                    and the Plaine &amp; Easie Code does not dictate the use of one method or the other.
+                </p>
+            </aside>
+            <aside class="example" title="Encoding Beams">
```
lpugin, line 1380: This does look right to me

ahankinson:
I'm not sure what happened here, but I think the best thing to do is close this one and re-open with a new PR

PR #156, opened 2025-12-04
Fixed: Clarify that empty bar lines (measures) should not appear.

### LINKED ISSUE #137: Clarify that empty bars are not permitted.
e.g., You can't do `//////`

ahankinson:
Also removes the SHOULD statement in response to feedback.

fixes #137

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1839,12 +1839,11 @@
         <section id="bar-lines">
             <h4>Bar Lines</h4>
             <p>
-                Bar lines serve to divide a musical staff into <a href="#measures">measures</a>.
-                Empty measures (two consecutive bar lines) are not permitted and MUST NOT be
-                present in the encoding.
+                Bar lines are indicated using one of the code options given in [[[#barlines-spec]]].
             </p>
             <p>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1816,8 +1816,8 @@ <h3>Measure and Staff Symbols</h3>
             <h4>Bar Lines</h4>
             <p>
                 Bar lines serve to divide a musical staff into <a href="#measures">measures</a>.
-                Empty measures&mdash;two consecutive bar lines with no musical notation
-                between them&mdash;MUST NOT appear in the encoding.
+                Empty measures (two consecutive bar lines) are not permitted and MUST NOT be
```
lpugin: I think this is problematic because two consecutive bar lines are permitted (as a double bar line)

ahankinson: They can't encode an empty measure with `//` since that MUST always be interpreted as a double bar line. This rule is to prevent `///` and `////`, etc.

ahankinson: It also removes any ambiguity that `//` is two single bar lines and an empty measure. Since empty bars (measures) cannot appear, then there is no way to interpret that as anything other than a single (double) bar line.

PR #157, opened 2025-12-04
Fixed: Add a section on ligatures

### LINKED ISSUE #28: Use of `u` to encode ligatures

ahankinson:
Fixes #28
Replaces #130

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1900,10 +1900,11 @@
         <section id="bar-lines">
             <h4>Bar Lines</h4>
             <p>
-                Bar lines MUST be indicated using one of the code options given in [[[#barlines-spec]]].
+                Bar lines are indicated using one of the code options given in [[[#barlines-spec]]].
             </p>
             <p>
-                Bar lines SHOULD be inserted at intervals that correspond to the current time signature.
+                Empty measures (two consecutive bar lines) are not permitted and MUST NOT be
+                present in the encoding.
             </p>
             <p>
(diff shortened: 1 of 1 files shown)


ahankinson:
No, and I purposely left it out because it's a bit of a grey area and not something we've discussed. If you want to encode a source in "modern" notation but keep the grouping of ligatures from the original (i.e., draw a bracket over them), then should that be allowed? 

We probably want it for neume notation as well.

I can open another issue if you want to address that specifically.

lpugin:
This section assumes that ligatures can occur in CMN, so we need to wait for #166 before we can move on with this one.

ahankinson:
Not really. I would prefer to merge this (it's the last major missing section) and then edit it later to specifically address #166. (We can't "release" the spec without dealing with all outstanding issues, so it won't get missed.)

ahankinson:
As for people using them for other things: we can’t reasonably police that all the time, but if we, for example, draw ligatures in CMN with square brackets, it’s much less likely that they would use it for a slur.

ahankinson:
I've copied your comment into #166. I think discussion should continue over there. Thanks!

PR #162, opened 2025-12-05
Update change log for ties

ahankinson:
Updates the change log to specify changes to tie encoding introduced in #155


ahankinson:
This should only be merged when #155 is merged. I'm just working through the backlog...

PR #169, opened 2026-01-06
WIP: Whole document review

ahankinson:
The changes here can be reviewed but the whole PR is a WIP as I do other passes through the document.

### DIFF
--- a/v2/guidelines.html
+++ b/v2/guidelines.html
@@ -356,6 +356,10 @@
                 </p>
             </section>
             <section>
+                Missing
+                accidentals in the original source SHOULD be supplied by the <a>transcriber</a>.
+            </section>
+            <section>
                 <h3>Tremolo, Slash</h3>
                 <p>
                     Notation abbreviations, such as tremolo, slash, etc., must
--- a/v2/guidelines.html
+++ b/v2/guidelines.html
@@ -302,6 +302,13 @@
             <section>
                 <h3>Chords</h3>
                 <p>
+                    Chords should be used sparingly, and only when they serve as a critical
+                    component for identifying a musical work. All else being equal, cataloguers
+                    should favour encoding a single voice since this increases the likelihood of
+                    positively matching incipits to each other when searching and comparing musical
+                    sources.
+                </p>
+                <p>
                     Encoding chords has changed considerably with Version 2.
                     Previously, chords were indicated by chaining together notes with
                     the <code>^</code> character. In Version 2, notes are grouped
--- a/v2/guidelines.html
+++ b/v2/guidelines.html
@@ -320,6 +320,14 @@
                 </p>
             </section>
             <section>
+                <h3>Neasure Rests</h3>
+                <p>
+                    Measure rests can be used to indicate the number of measures that are being skipped in the
+                    <a>encoding</a> before the musical content being captured by the incipit, regardless of whether
+                    these measures in the <a>original source</a> contain musical content.
+                </p>
+            </section>
+            <section>
                 <h3>Neume and Mensural Notation</h3>
                 <p>
                     It is possible to change the specified notation system to neume and
--- a/v2/index.html
+++ b/v2/index.html
@@ -1693,13 +1693,12 @@
             </p>
             <p>
                A <a>duration</a> value MAY precede the <code>^</code>. Two or more <a>note names</a>
-               (with optional octave and accidental indications) MUST immediately follow this character.
-               The notes MUST be ordered from the lowest to the highest on the staff line. A <code>></code>
-               MUST be used to end the chord. A duration value for the individual notes within a chord MUST
-               NOT be supplied.
+                (with optional octave and accidental indications) MUST immediately follow the <code>^</code>.
+               Notes within the chord SHOULD be ordered from the lowest to the highest on the staff line.
+               Duration values for the individual notes within a chord MUST NOT be supplied.
             </p>
             <p>
-                If the duration is omitted, the last specified duration is used. If no duration is
+                If the chord duration is omitted, the last specified duration is used. If no duration is
                 supplied in the <a>encoding</a> a default duration of <code>4</code> is assumed.
             </p>
             <p>
--- a/v2/index.html
+++ b/v2/index.html
@@ -126,7 +126,6 @@
     </style>
 </head>
 <body>
-<div class="issue" data-number="31"></div>
 <p class="copyright">
     This document is licensed under a
     <a class="subfoot" href="https://creativecommons.org/licenses/by/4.0/" rel="license">
@@ -277,7 +276,6 @@
     </p>
     <section id="clef">
         <div class="issue" data-number="72"></div>
-        <div class="issue" data-number="119"></div>
         <h3>Clef</h3>
         <p>
             Every encoding MUST include a clef.
@@ -1367,7 +1365,6 @@
     <section id="groups">
         <h3>Groups</h3>
         <section id="ties">
-            <div class="issue" data-number="112"></div>
             <h4>Ties</h4>
             <p>
                 Tied notes are indicated with an underscore character, <code>_</code>. Tied notes consist of
@@ -1455,7 +1452,6 @@
             </aside>
         </section>
         <section id="ligatures">
-            <div class="issue" data-number="28"></div>
             <h4>Ligatures</h4>
             <p>
                 Two notes joined in a ligature can be indicated with the lower-case character
@@ -1684,9 +1680,6 @@
         </section>
         <section id="chords">
             <h4>Chords</h4>
-            <div class="issue" data-number="2"></div>
-            <div class="issue" data-number="60"></div>
-            <div class="issue" data-number="65"></div>
             <p>
                 Notes in a chords are enclosed in the characters <code>^</code> and <code>></code>. All
                 notes in a chord are assigned the same duration value.
--- a/v2/index.html
+++ b/v2/index.html
@@ -206,7 +206,7 @@
         <dt><dfn>Common Western Music Notation</dfn></dt>
         <dd>
             A system of written music notation largely formalized in Europe in the 17th century and widely
-            practiced today. It features staves, measures, and chords among other musical notation figures.
+            practiced today.
         </dd>
         <dt><dfn>Encoding</dfn></dt>
         <dd>
--- a/v2/index.html
+++ b/v2/index.html
@@ -225,7 +225,7 @@
         <dt><dfn>Mensural Music Notation</dfn></dt>
         <dd>
             A system of written musical notation, distinguished by its use of proportion-based durations and
-            lacking measures.
+            no measures.
         </dd>
         <dt><dfn>Neume Notation</dfn></dt>
         <dd>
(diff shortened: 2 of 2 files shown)


v2/index.html
```diff
@@ -189,23 +193,83 @@ <h2>Introduction</h2>
...
+        <dt><dfn>Mensural Music Notation</dfn></dt>
+        <dd>
+            A system of written musical notation, distinguished by its use of proportion-based durations and
+            lacking measures.
+        </dd>
+        <dt><dfn>Neume Notation</dfn></dt>
+        <dd>
+            A system of written musical notation, widely used in Western countries representing
+            the musical notation for a sung liturgical text, e.g., "plainchant". This notation uses
+            neumes: Shapes which convey precise pitch information, have no features for capturing
```
```suggestion
            neumes: Shapes which convey pitch information, have no features for capturing
```

v2/index.html
```diff
@@ -357,40 +421,28 @@ <h3>Key Signature</h3>
...
-            note names are not in the original source and have been supplied by the <a>transcriber</a>. Consecutively
+            note names have been supplied by the <a>transcriber</a> and are not in the original source. Consecutively
             supplied note names MUST be within a single set of brackets. A key signature MAY contain more than one
             set of non-consecutive bracket groups.
         </p>
         <p>
-            A key signature containing a single <code>n</code> character MAY be supplied to indicate a natural key
-            signature, e.g., C Major or A minor. This character MUST NOT be followed by any note names.
+            A key signature, including the <code>x</code>, <code>b</code>, and <code>n</code> characters MAY be
+            entirely enclosed in square brackets, meaning the cataloguer has supplied the full key signature.
```
lpugin: I think it would be better not to include the accidental type in the `[]`. That is allowing `x[FC]` but not `[xFC]`, even though the meaning is the same.

v2/index.html
```diff
@@ -484,37 +536,33 @@ <h3>Time Signature</h3>
...
         </p>
         <p>
             <abbr title="Common Western Music Notation">CWMN</abbr> time signatures are expressed as one number
-            above another. These numbers MUST be separated by a <code>/</code> character. Any positive digit MAY
-            be used, but <a>encoders</a> SHOULD use commonly accepted values where possible. The <code>c</code>
-            ("common", or <code>4/4</code>) and <code>c/</code> ("alla breve", or <code>2/2</code>) characters
-            MAY be used.
+            above another. These numbers are separated by a <code>/</code> character. Any positive digit MAY
+            be used, but <a>transcribers</a> SHOULD use commonly accepted values where possible. The <code>c</code>
+            ("common") and <code>c/</code> ("alla breve") characters MAY be used.
```
lpugin: We should clarify that `c6/8` is allowed

v2/index.html
```diff
@@ -484,37 +536,33 @@ <h3>Time Signature</h3>
...
-            tempus, respectively. The <code>.</code> character indicates "major" prolation; omitting <code>.</code>
-            indicates "minor" prolation. A <code>/</code> character may follow the <em>tempus</em> character to
-            indicate diminution.
+            tempus, respectively. A <code>/</code> character can follow the <em>tempus</em> character to
+            indicate diminution. The <code>.</code> character indicates "major" prolation; omitting <code>.</code>
+            indicates "minor" prolation.
         </p>
         <p>
-            A mensuration sign MAY include a numerical component as a proportion or augmentation,
+            A mensuration sign can include a numerical component as a proportion or augmentation,
```
lpugin: Proportions can be different from 3 and 2, so I am not sure where the limitation came in before this PR.

v2/index.html
```diff
@@ -691,16 +739,18 @@ <h2>Musical Notation</h2>
...
-            of a note or rest. Complex notational figures, such as chords, beams, or tuplets, serve
-            to <a>group</a> notes and rests. <a>Expression Marks</a> alter the performance of a note, a rest, or
-            groups thereof. Other <a>staff and measure symbols</a> can control the definition of the staff
-            itself: bar lines, cross-measure rests, or a change of clef, key signature, or time signature.
+            a <a>staff</a> of musical notation. <a href="#notes-and-rests">Notes and rests</a>, and chords,
+            are the most basic <a>logical units</a> of notation representation, composed of one or more
+            characters that specify different attributes of a note or rest. Complex notational figures,
+            such as beams, or tuplets, serve to group these logical units. <a href="#expression-marks">Expression marks</a>
+            alter the performance of a note, a rest, or groups thereof. Other staff and <a>measure</a> symbols
+            can control the definition of the staff itself: bar lines, cross-measure rests, or a change of clef,
```
```suggestion
            can control the definition of the staff itself: bar lines, multi-measure rests, or a change of clef,
```

v2/index.html
```diff
@@ -1233,32 +1284,41 @@ <h4>Accidentals</h4>
...
-                Name</a> until the next <a href="#bar-lines">Bar Line</a>. If the same accidental is given to several
-                notes with the same <a>Note Name</a> within the bar, it will have no further effect.
+                Where a note in a <a>key signature</a> is altered by use of an
+                accidental directly before the note, the <a>written pitch</a> and
+                <a>sounding pitch</a> of the note is that indicated by the accidental.
             </p>
             <p>
-                Accidentals MUST be interpreted by their written value, and MUST NOT be interpreted by their values
-                relative to a preceding accidental.
+                A note cannot be altered by the same accidental more than once within the same bar. The
```
```suggestion
                A note cannot be altered by the same accidental more than once within the same measure. The
```

v2/index.html
```diff
@@ -1233,32 +1284,41 @@ <h4>Accidentals</h4>
...
+                alteration of a note continues on subsequent notes with the same <a>note name</a> until the
+                next <a href="#bar-lines">bar line</a>. If an accidental is given to subsequent notes
+                with the same note name in the same measure, it has no effect.
```
lpugin: I don't think this is necessary. Nobody would expect `xCxCxC` to have each note going up one more semi-tone - if this is what is being described here.

ahankinson:
🤷 We could go either way. I chose `and` because that seems to be what it was called originally, but I'm just as happy with `&`.

v2/index.html
```diff
@@ -2060,21 +2137,22 @@ <h4>Bar Lines</h4>
...
+                measures in this system of notation.
             </p>
-            <p>
-                Measure rests MAY indicate the number of measures that are being skipped in the
-                <a>original source</a> before the musical content being captured by the incipit, regardless of whether
-                these measures in the original source contain musical content.
+            <p class="note">
+                Measure rests can be used to indicate the number of measures that are being skipped in the
+                <a>encoding</a> before the musical content being captured by the incipit, regardless of whether
+                these measures in the <a>original source</a> contain musical content.
```
lpugin: More of a guideline statement to me.

v2/index.html
ahankinson: I think the intention is that `xCxC` would not show the sharp symbol twice...

lpugin: I don't agree with that. If you don't want it to be shown twice, you encode `xCC` - at least that is how Verovio works now.

v2/index.html
```diff
@@ -189,23 +193,83 @@ <h2>Introduction</h2>
...
+            MIDI.
+        </dd>
+        <dt><dfn>Sounding Pitch</dfn></dt>
+        <dd>
+            The pitch of the instrument when it is performed, independent of how the
+            notation looks for that instrument on a musical staff.
+        </dd>
+        <dt><dfn>Staff</dfn></dt>
+        <dd>
+            Five parallel lines that serve to indicate the pitch of a note relative to
```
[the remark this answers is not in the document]
lpugin: OK

v2/index.html
ahankinson: What I'm trying to say is that `xCC` and `xCxC` are equivalent encodings... they mean the same thing. Whether or not a renderer chooses to show:

image

or

image

is not prescribed by the specification -- both renderings are correct.

v2/index.html
ahankinson: What does that mean?

lpugin: It usually means a 6/8 where the c (or c/) conveys a tempo change indication (e.g., twice as fast). Not very common, but definitely to be found in 17th century sources.

v2/index.html
lpugin: OK. Then I think the previous sentence is good enough, and that that one is introducing confusion more than anything else

v2/index.html
```diff
@@ -1233,32 +1284,41 @@ <h4>Accidentals</h4>
...
             <aside class="note">
                 <p>
-                    Accidentals are interpreted according to their written value. They do not have a cumulative effect
-                    if combined; that is, specifying a flat accidental on a note that is sharpened does not lower the
-                    pitch by a semitone.
+                    Accidentals follow the established convention in music notation writing that alterations
+                    to a note within a measure do not require the accidental to be restated. This is also reflected
+                    in the encoding, where notes are not required to restate the accidental for subsequent notes in
+                    a measure with the same note name. <a>Renderers</a> can choose to draw the repeated
+                    accidentals, or they can choose to render only the first.
```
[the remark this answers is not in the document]
lpugin: I would drop this sentence because if we have some renderers that start not showing repeated accidentals, it would start giving the impression that accidentals have to be repeated for the sounding pitch to be alterated.

v2/index.html
lpugin: I would drop it. The following sentence is better and enough.

v2/index.html
```diff
@@ -1233,32 +1284,41 @@ <h4>Accidentals</h4>
...
-                notes with the same <a>Note Name</a> within the bar, it will have no further effect.
+                Where a note in a <a>key signature</a> is altered by use of an
+                accidental directly before the note, the <a>written pitch</a> and
+                <a>sounding pitch</a> of the note is that indicated by the accidental.
             </p>
             <p>
-                Accidentals MUST be interpreted by their written value, and MUST NOT be interpreted by their values
-                relative to a preceding accidental.
+                A note cannot be altered by the same accidental more than once within the same bar. The
+                alteration of a note continues on subsequent notes with the same <a>note name</a> until the
```
```suggestion
                alteration of a note continues on subsequent notes with the same <a>note name</a> and with the same octave until the
```

v2/index.html
ahankinson: It looks like `c6/8` isn't currently supported, so that's a new feature. I've created issue #175 to track that, but I'll consider this resolved for now.

ahankinson:
OK; I have finished this review. I have tried to keep changes to only address clarity, so hopefully there are no surprises.

v2/guidelines.html
```diff
@@ -347,6 +362,8 @@ <h3>Measure rests</h3>
                     original source before the musical content being captured by the incipit, regardless of whether
                     these measures in the original source contain musical content.
                 </p>
+                Missing
```
lpugin, line 365: What do we mean by missing accidentals? Can we clarify? Otherwise there is a risk that catalogers will think they have to add every (sounding) accidental

ahankinson: Hm. I think I was going to take this out because it looked too much like an encoding guideline...  Would it be ok to remove this line?

ahankinson: Oh, wait... this is the guidelines! Nevermind. They haven't been edited *at all* yet, and should simply be read as a bunch of TODOs. I think I moved this here when I removed it from the spec, just so it didn't get forgotten.

lpugin: OK. I was not sure if it was new or not. We can leave it for now

lpugin (approved):

PR #177, opened 2026-05-11
Fixed: ligatures in CWMN

### LINKED ISSUE #166: Should ligatures be available in CMN encoding
#28 specifies that `u` can be used to encode ligatures. 

We should decide whether this is available only on Mensural or Neume notation, or available on all types of notation.

From #157 : 

> If you want to encode a source in "modern" notation but keep the grouping of ligatures from the original (i.e., draw a bracket over them), then should that be allowed?

ahankinson:
Clarifies that ligatures are not valid in CWMN encodings.

Fixes #166

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1463,8 +1463,7 @@
             </p>
             <p>
                 A ligature MUST NOT occur between a note and a rest, and MUST NOT occur between a note and a group
-                of notes. (i.e., a beam, tuplet, or chord.) Ligatures also MUST NOT occur within a beam, tuplet,
-                or chord.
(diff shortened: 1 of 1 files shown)


lpugin (approved):

lpugin (approved):

BaMikusi:
Sorry, one more belated comment on a PR already merged.

We have two statements:
(1) Ligatures can only be used in Mensural and Neume Notations, but NOT in CWMN.
(2) "Ligatures also MUST NOT occur within a tuplet."

Isn't (2) superfluous, once (1) has been declared? In other words, can you meaningfully encode tuplets in Mensural and Neume Notations?

PR #179, opened 2026-05-11
Fixed: Allow supplied in neume keysig

ahankinson:
Fixes #173

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -438,11 +438,11 @@
         </p>
         <p>
             A key signature, including the <code>x</code>, <code>b</code>, and <code>n</code> characters MAY be
-            entirely enclosed in square brackets, meaning the cataloger has supplied the full key signature.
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -456,11 +456,11 @@ <h3>Key Signature</h3>
         </p>
         <p>
             A key signature, including the <code>x</code>, <code>b</code>, and <code>n</code> characters MAY be
-            entirely enclosed in square brackets, meaning the cataloger has supplied the full key signature.
+            entirely enclosed in square brackets, meaning the transcriber has supplied the full key signature.
```
lpugin, line 459: I am still not convinced by the fact we can have fully enclosed key signatures.

ahankinson: That is addressed with #172. This PR doesn't change any rules to that effect.

lpugin (approved):

PR #184, opened 2026-05-12
Fixed: Further clarify octave and durations for beams

ahankinson:
Including the rules for how they affect nested appoggiatura groups.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1553,10 +1553,6 @@
                 duration values.
             </p>
             <p>
-                Duration and octave values in an appoggiatura group, including values that precede a nested beam,
-                apply only to that appoggiatura group and MUST NOT affect the outer beam group.
-            </p>
-            <p>
                 Notes with duration values of <code>0</code>, <code>1</code>, <code>2</code>,
                 <code>4</code>, and <code>9</code> MUST NOT occur within
                 a beam group. Notes with other duration values MAY occur within a beam group.
--- a/v2/index.html
+++ b/v2/index.html
@@ -1548,9 +1548,9 @@
             </p>
             <p>
                 A beam group MAY be immediately preceded by an octave indication, a duration value,
-                or both. When present, these values apply to the note or rest in the beam group
-                and remain in effect for the notes or rests in the beam until a new value is provided, following the
-                <a href="#implied-duration-octave">convention</a> for implied octave and duration values.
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -837,10 +837,10 @@ <h3>Structure</h3>
...
             time signature, from the music notation that follows the change.
         </p>
-        <p>
-            To aid in writing succinct incipits, <a href="#durations">duration</a> and <a href="#octaves">octave</a> values do not need to be
-            stated on every note. Where they are absent, the previously stated value is used. Where no values
-            are stated in the encoding, a default value is usually inferred. See the sections on
+        <p id="implied-duration-octave">
+            To aid in writing succinct incipits, <a href="#durations">duration</a> and <a href="#octaves">octave</a>
+            values do not need to be stated on every note. Where they are absent, the previously stated value is implied.
+            Where no values are stated in the encoding, a default value is usually inferred. See the sections on
```
lpugin: Not sure what is meant by "usually inferred" ? Why usually and what should it be inferred from? Shouldn't we say "can be decided"? or "can be assumed"?

v2/index.html
```diff
@@ -1545,6 +1545,16 @@ <h4>Beams</h4>
...
                 <a href="#appoggiatura-groups">appoggiatura group</a>.
             </p>
+            <p>
+                A beam group MAY be immediately preceded by an octave indication, a duration value,
+                or both. When present, these values apply to the note or rest in the beam group
+                and remain in effect for the notes or rests in the beam until a new value is provided, following the
+                <a href="#implied-duration-octave">convention</a> for implied octave and duration values.
+            </p>
+            <p>
+                Duration and octave values in an appoggiatura group, including values that precede a nested beam,
```
lpugin: That has never been the case - I think.

ahankinson: It was worth mentioning, though, since it doesn't follow the rules of previously-stated durations and octaves.

v2/index.html
ahankinson: I've tightened up this prose.

v2/index.html
lpugin: I mean, this is not we currently do. A duration change in the appogiatura group remains in effect at the end of (and after) the group. I don't think we want to change this, or have I missed something?

ahankinson: I've removed the paragraph.

lpugin (approved):

Issue #132, opened 2024-06-05
Add a note clarifying the potential change of the repetition context.


lpugin:
It would be good to make users aware that when using repetitions (groups or measures), the context when the repetition code is being applied might be different than when the group being repeated started. This is certainly important for octaves and durations. Any thing else?

In other words, transcriber needs to keep in mind that the repetitions repeat the code as is and not the notation represented. Or at least, depending on the context, the result might be different. (We probably need an example...)

ahankinson:
Depending on how the repeated rhythm sequences land, it could also apply — if you have a set of durations and then the repeat group only covers a subset of them.

BaMikusi:
Sorry, I just see this new issue -- so then I reference the example I just entered elsewhere:
https://github.com/rism-digital/pae-code-spec/pull/110#issuecomment-2150742412

lpugin:
Thinking about this overnight I think we need to evaluate if repetitions indicate a repetition of the code and not of the music is really what was originally meant. It might be just a side effect of renderers like Verovio. It seems to me that reversing this would not be problematic because to have it work with the current behaviour of Verovio, transcribers had to make the code more explicit That is, adding additional context duration and / or octave at the beginning of the group to be repeated. 

This means that we could re-evaluate this and change Verovio accordingly without having problems with the old data. Actually, it might be even fixing old data that we haven't identified as problematic because of this. In the end, this would be not only a clarification, but also an improvement for transcribers since the change context side effect in the use of the repetitions always comes "as a surprise" when cataloguing.

lpugin:
PS another context that might have changed in the course of a repetition declaration is the clef, but since this is visually given in the notation, I don't think it would be necessary - or even correct? to reset it when the repetition is applied. Any thoughts?

ahankinson:
Depends on how you define the code and its interpretation. If you simply say something like "When segments are repeated, the context is taken from the initial declaration and not from the repeat" then that provides the rule in how the code should be evaluated. (The wording would need to be finessed, of course.)

So the code is always correct, if you first define what "correct" is. :-D

BaMikusi:
Fair enough.

As for @lpugin' other question regarding clefs, I think these should also be implied on the repeat, just as durations and octaves. This, for example, is not an unrealistic scenario:
8,C%G-2 'G^E %F-4 ,,B %G-2 'G^F^D/i/
[...]

PR #186, opened 2026-05-12
Fixed: shortcuts section

ahankinson:
Fixes and normalizes the shortcuts section. This is intended primarily as a clarification, not as introducing any new rules.

The rhythmic repeat was moved as a sub-section to the Duration section. Since this can apply in several places, it made more sense to define it as a variation of duration encoding, rather than its own standalone shortcut. It also introduces no new characters for encoding, and instead offers an alternative to how the durations can be applied. It also simplifies references to this, as any logical unit that refers to a duration can also be assumed to allow rhythmic sequences as well.

Also clarifies #132

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1114,18 +1114,18 @@
                     and continues until the duration pattern is ended.
                 </p>
                 <p>
-                    A duration pattern ends when a new duration value or duration pattern is supplied. It also
-                    ends at a <a href="#changes-to-staff-definitions">staff definition change</a>,
-                    <a href="#tuplets">tuplet boundary</a>, or <a href="#repeat-group">repeat group</a>.
+                    A duration pattern ends when a new duration value or duration pattern is supplied.
                 </p>
                 <p>
-                    Duration patterns MUST NOT be used with tuplets, since they provide their own internal durations.
+                    When applied to tuplets, duration patterns affect the written duration of the notes, and not
+                    the musical duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
+                    more information.
                 </p>
                 <p>
                     <a href="#ties">Tied notes</a> use two duration values in the duration pattern.
                 </p>
                 <p>
-                    Duration pattern shortcuts MUST NOT be used with Neume notation, since notes in Neume
+                    Duration patterns MUST NOT be used with Neume notation, since notes in Neume
                     notation do not have duration values.
                 </p>
                 <aside class="example" title="Encoding Duration Patterns">
@@ -1152,6 +1152,22 @@
                             <td class="notation-result"></td>
                             <td>
                                 Instead of <code>{'8.A6B''8C}{8.D6E8F}</code> the code can be <code>'8.68{AB''C}{DEF}</code>.
+                            </td>
+                        </tr>
+                        <tr class="notation-example">
+                            <td class="notation-code">
+                                <script type="application/json">
+                                    {
+                                        "clef": "G-2",
+                                        "timesig": "3/2",
+                                        "data": "'8.68(AB''C)(DEF)"
+                                    }
+                                </script>
+                                <code>'8.68(AB''C)(DEF)</code>
+                            </td>
+                            <td class="notation-result"></td>
+                            <td>
+                                Duration patterns applied to tuplets.
                             </td>
                         </tr>
                         </tbody>
--- a/v2/index.html
+++ b/v2/index.html
@@ -2614,7 +2614,7 @@
             <h5 id="measure-repeat">Measure repeat</h5>
             <p>
                 A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
-                represents one repetition of the most recent complete measure that is not itself a measure
+                represents one repetition of the most recent complete measure
                 repeat.
             </p>
             <p>
--- a/v2/index.html
+++ b/v2/index.html
@@ -2354,23 +2354,24 @@
             </aside>
         </section>
         <section>
-            <h5>Repeated measures</h5>
+            <h5 id="measure-repeat">Measure repeat</h5>
             <p>
-                If one or more whole measures are repeated, a measure
-                repeat can be used.
+                A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
+                represents one repetition of the most recent complete measure that is not itself a measure
+                repeat.
             </p>
             <p>
-                The measure to be repeated MUST end with a bar line. The lower-case
-                character <code>i</code> MUST occur immediately after this bar line,
-                and MUST be immediately followed by another bar line. There MUST NOT
-                be any other characters between these two bar lines.
+                The measure to be repeated MUST end with a bar line. The lower-case character <code>i</code>
+                MUST occur immediately after this bar line and MUST be immediately followed by another bar
+                line. There MUST NOT be any other characters between <code>i</code>
+                and the surrounding bar lines.
             </p>
             <p>
-                Any type of bar line MAY be used to begin or end a measure repeat group.
+                Any type of <a href="#bar-lines">bar line</a> MAY be used to begin or end a measure repeat.
             </p>
             <p>
-                Measure repeats SHOULD NOT be used with Mensural or Neume notation due to the
-                general absence of measures in this system of notation.
+                Measure repeats SHOULD NOT be used with Mensural or Neume notation due to the general
+                absence of measures in these notation types.
             </p>
             <aside class="example" title="Encoding Repeated Measures">
                 <table class="simple" style="width: 100%">
--- a/v2/index.html
+++ b/v2/index.html
@@ -2557,10 +2557,9 @@
             </p>
             <p>
                 To encode a repeat group, mark the beginning and end of the repeated figure with an
-                exclamation mark character <code>!</code>. The lower-case character <code>f</code> MUST
-                immediately follow the closing <code>!</code>. The number of <code>f</code> characters
-                indicates the number of times the figure is repeated. At least one <code>f</code> character
-                MUST be supplied.
+                exclamation mark character <code>!</code>. At least one lower-case character <code>f</code>
+                MUST immediately follow the closing <code>!</code>. The number of <code>f</code> characters
+                indicates the number of times the figure is repeated.
             </p>
             <p>
                 The repeated figure and all of its repetitions MUST occur within the same measure. However,
--- a/v2/index.html
+++ b/v2/index.html
@@ -2279,6 +2279,12 @@
     </section>
     <section id="shortcuts">
         <h3>Shortcuts</h3>
+        <p>
+            Shortcuts provide compact ways to encode repeated musical material. A shortcut represents
+            notation that could also be written out in full. Unless otherwise stated, shortcut syntax
+            follows the same rules as the notation it replaces, including the conventions for implied
+            duration and octave values.
+        </p>
         <section id="repeat-group">
             <h4>Repeat group</h4>
             <p>
--- a/v2/index.html
+++ b/v2/index.html
@@ -2282,18 +2282,28 @@
         <section id="repeat-group">
             <h4>Repeat group</h4>
             <p>
-                If one or more notes or rests are repeated several times, a repeat group
-                can be used.
+                A repeat group encodes a figure that is repeated within a single measure.
             </p>
             <p>
-                A repeat group is only valid within a single measure.
+                A repeat group MUST NOT contain a bar line, measure rest, measure repeat, staff definition
+                change, or another repeat group.
             </p>
             <p>
-                To use a repeat group, mark the beginning and the end of the group with
-                an exclamation mark character <code>!</code>. The lower-case character
-                <code>f</code> MUST immediately follow the ending <code>!</code>, and
-                MUST occur the number of times the figure should be repeated. The
-                <code>f</code> MUST be specified at least once.
+                To encode a repeat group, mark the beginning and end of the repeated figure with an
+                exclamation mark character <code>!</code>. The lower-case character <code>f</code> MUST
+                immediately follow the closing <code>!</code>. The number of <code>f</code> characters
+                indicates the number of times the figure is repeated. At least one <code>f</code> character
+                MUST be supplied.
+            </p>
+            <p>
+                When a repeat group is expanded, each repetition MUST preserve the musical values represented by
+                the original repeated figure. Implied duration and octave values within the repeated figure MUST be
+                resolved from the context at the start of the repeat group for each repetition, not from the end of
+                the previous repetition.
+            </p>
+            <p>
+                The repeated figure and all of its repetitions MUST occur within the same measure. However,
+                a repeat group MAY be combined with a <a href="#measure-repeat">measure repeat</a>.
             </p>
             <aside class="example" title="Encoding Repeated Figures">
                 <table class="simple" style="width: 100%">
@@ -2318,6 +2328,20 @@
                         </td>
                         <td class="notation-result"></td>
                         <td>Repeat twice</td>
+                    </tr>
+                    <tr class="notation-example">
+                        <td class="notation-code">
+                            <script type="application/json">
+                                {
+                                    "clef": "G-2",
+                                    "timesig": "3/2",
+                                    "data": "!8{AB}!f/i/i"
+                                }
+                            </script>
+                            <code>!8{AB}!f/i/i</code>
+                        </td>
+                        <td class="notation-result"></td>
+                        <td>Repeat the figure, then repeat the measure.</td>
                     </tr>
                     </tbody>
                 </table>
--- a/v2/index.html
+++ b/v2/index.html
@@ -1148,7 +1148,7 @@
                 </p>
                 <p>
                     When applied to tuplets, duration patterns affect the written duration of the notes, and not
-                    the musical duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
+                    the <a>sounding duration</a> of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
                     more information.
                 </p>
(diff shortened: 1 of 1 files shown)


v2/index.html
rism-international, line 1: For line 1807:

Is there a specific reason for not allowing a rhythmic sequence to run beyond a bar line? Once e.g. an accompaniment pattern begins, it could frequently go on for a while, irrespective of bar lines, so making transcribers write out the rhythm once again for each measure may be superfluos hassle for them?

And BTW: Could the term "rhythmic pattern" prove helpful in describing this phenomenon overall? "Rhythmic sequence" is okay, but when analyzing such a passage it is rather the term "pattern" that first comes to my mind (as it has spontaneously happened in the above paragraph).

lpugin: Rhythmic patterns currently do not end at the end of a measure, beam, tuplet, etc. They only end when a new duration value (or new rhythmic pattern) occurs. I don't think we want to change that.

I agree that "pattern" is better than "sequence"

ahankinson: Looking at the PR I just closed, I can see that I had previously named this a "Duration Group", which I tend to like a bit better since we tend to call these things "Durations" and don't use the term "Rhythmic" anywhere else in the spec. Or possibly "Duration Pattern"?

lpugin: Duration pattern is good for me

lpugin: (even though I would find it fair to claim that the notion of "rhythm" emerges only once you have more than one duration, so rhythmic pattern would be fine for me even though we don't use anywhere else)

ahankinson: Yes; I guess it depends on whether you think you are supplying a rhythm, or you're using a shortcut to avoid needing to supply individual durations to the notes. I tend to think it's the latter?

BaMikusi: I appreciate the aesthetic (as well as logical) value of not introducing at this point a category that is otherwise absent from the specs, so "duration pattern" is fine.

v2/index.html
```diff
@@ -1100,6 +1100,64 @@ <h4>Durations</h4>
...
+                <p>
+                    A duration pattern consists of two or more duration values followed by a sequence of notes,
+                    rests, or chords. The duration values are assigned to those logical units in order. Each note,
+                    rest, or chord uses one duration value from the pattern. If the sequence contains more logical
+                    units than duration values, the duration pattern applies cyclically from its first duration value
+                    and continues until the duration pattern is ended.
+                </p>
+                <p>
+                    A duration pattern ends when a new duration value or duration pattern is supplied. It also
+                    ends at a <a href="#changes-to-staff-definitions">staff definition change</a>,
```
lpugin: That's never been the case, and I think it is problematic to introduce this change. For instance, why is a clef change not allowed? And also, which duration is in force after? The last one used in the pattern? The next one? The first one of the pattern? That is ambiguous.

v2/index.html
```diff
@@ -1100,6 +1100,64 @@ <h4>Durations</h4>
...
+                    <a href="#tuplets">tuplet boundary</a>, or <a href="#repeat-group">repeat group</a>.
+                </p>
+                <p>
+                    Duration patterns MUST NOT be used with tuplets, since they provide their own internal durations.
```
lpugin: Same here, it has never been the case (and never a problem). Also, I don't think it is right to say that tuplets provide their own internal durations. You can do
`8CC(CCC)CC`. And might want to do `8CC8.68(CCC)(CCC)`

v2/index.html
ahankinson: 1. It may have never been the case in Verovio, but it’s never been written down. There are a lot of rules that were never written down that need to be clarified. 
2. staff definition changes can also change the time signature. I think it’s better to provide a clear “stop repeat” point, and that seemed like a logical one to add.
3. I would expect the normal rules of implied duration would take over; that is, re-state a new duration or use the quarter-note implied duration.

v2/index.html
ahankinson: Again, it may never have been the case in Verovio, but this was never written down so I am trying to write it down.

tuplets absolutely do provide an internal duration since the tuplets you provided are 3 in the space of 2. They may have the visual appearance of an eighth note but their “outer” duration is a quarter note.

I thought it best not to confuse the matter and simply exclude them from duration patterns entirely.

ahankinson: I’m not writing these from the perspective of simply formalizing what Verovio does. I am writing them trying to balance expressiveness with simplicity of parsing and representation. There is very little to go on in the original papers (nothing, in fact) and a single sentence in V1.

lpugin: Well, tuplets provide internal duration **proportions**, so if you want to say so, you should be explicit about that because otherwise it is confusing. But that is a side point here, the main one being why excluding pattern from tuplets. Can you explain why? The example `8CC8.68(CCC)(CCC)` is a very realistic one where, as a cataloger, I would want to use patterns.

lpugin: Yes, of course, Verovio is in no way driving what V2 should be. I am only raising the fact that is have never been a problem (in and outside Verovio), so if we introduce a restriction, there should be a clear reason about it, and we also need to be sure it does not introduce an ambiguity we did not have before.

ahankinson: Because currently 8.68(CCC)(CCC) is not going to be valid, since we’re talking about removing the outer duration value, and only having the “written” duration value **inside** of the parentheses. So it’s inconsistent to specify a single duration inside the parens, but multiple durations outside.

And if we do move the total duration value inside (e.g., `,4` or whatever) it will be at the end inside the parentheses, making its application in a repeat group problematic.

ahankinson: That’s what I meant by tuplets having their own duration “logic” since the note “duration shapes” are  stated inside the parentheses, which is incompatible (to my mind) with duration patterns.

And as written the first character inside a tuplet needs to be a duration: “ The first encoded note, rest, chord, or beam group within a tuplet MUST include an explicit duration value…”

lpugin: I appreciate the effort in clarifying things, and I am aware that this can come with some new restrictions. But if `8.68(CCC)(CCC)` is not valid, then I feel we are moving away from the idea of plain and easy, and that is a big problem for me. So having the restriction that "a tuplet MUST include an explicit duration value" looks like a restriction we should not introduce. I do not understand why we need it, and I do not understand where the problem is with the example, assuming that the optional full tuplet duration is inside - meaning you could do `8.68(4:CCC)(CCC)` if you want.
(We don't have to continue the discussion here if you feel this is not appropriate)

ahankinson: Duration groups are an “advanced” feature that has been under specified for 60-odd years, and to my mind is not really central to being “plain” and “easy” (most cataloguers probably don’t use it and they get by just fine). So I’m skeptical of the claim that restricting their interpretation in the specific case of tuplets (themselves also under-specified) where duration interpretation is already tricky, qualifies as a reason to consider that we are wholesale abandoning those principles. 

But to speak to the specifics, there would be three ways to express 8.68(CCC) (leaving aside repetitions for the moment), if we consider that you can specify duration BOTH inside and outside the tuplet (which is what you are suggesting, as I understand it): 8.(C6C8C), AND (8.C6C8C). All three would result in the same notation. What this rule is saying is that only the last one is valid, since the note values for a tuplet are only valid if they are defined inside the tuplet.

Adjusting the tuplet requirements should be noted on #195 so it doesn’t get lost here.

v2/index.html
BaMikusi: I am also skeptical as to forcing the duration pattern to end whenever a new staff definition appears. A clef change is just a matter of convenience that does not the least imply that we have a musical caesura on our hands. And even if the time signature itself changes, that may or not have an influence on the pattern. If it does not, and the rhythm can be notated the same way as before, there is no need to stop the pattern, either -- and if this is not the case (which would be the majority), the rhythm is surely different anyhow and so the temptation to run on with the pattern is just not there anyway. Overall, a new staff definition very often implies no substantial break in the musical texture, and should thus not necessarily be viewed as a breaking point in the code, either -- by which I am also saying that relying at such a point on implied duration would seem quite problematic to me; that should (I think) only be an option at the very start of the encoding.

v2/index.html
BaMikusi: Verständnisfrage: Would (8.68CCC) not count as a valid alternative spelling of (8.C6C8C)?

lpugin: It is interesting you see duration patterns as an "advanced" feature. From my (little) experience from typesetting incipits, patterns make it easier, simpler and less error prone. Take the very common pattern:
```
(8.C6D8C)(8.D6E8D)(8.C6C8C)(8.D6E8D)
```
versus
```
8.68(CDC)(DED)(CDC)(DED)
```
The second is much simpler to typeset because the encoder only need to focus on note pitches. So as an encoder, I think I would be quite annoyed if patterns are not allowed in tuplets in PAE2, and I would not assume that cataloguers don't use it.

ahankinson: Possibly? 

But there's not much point in doing that either, since it would only apply within the tuplet, and not repeated to the next tuplet. My point was trying to illustrate that 8.68(CCC) didn't follow the rules for tuplets, since the outer duration (the one attached to the `(`) should have no effect on the shape of the inner notes, since that duration is defined as being the "total duration" (i.e., the musical duration) of the tuplet, not the written appearance of the notes in the tuplet. `4(8CCC)` is "three eighth notes in the span of a quarter" (at least that's how it's supposed to be right now). 

But I'm reviewing @lpugin's response so this may all change anyway.

ahankinson: > patterns make it easier, simpler and less error prone

I don't disagree with that! I just meant that it's a feature most 'casual' transcribers don't use regularly, and the rules for its application are not completely obvious. *Most* incipits where a duration group _could_ be used do not use them. 

I'm also not disputing that it's useful for tuplets. It's just that we have a chicken-and-egg problem at the moment, where duration interpretations for tuplets are under-specified, and that makes defining how duration patterns apply  to them, in a way that is consistent with the general principles of applying durations, not completely obvious.

v2/index.html
ahankinson: OK

v2/index.html
BaMikusi: To clarify, I was asking about the (8.68CCC) option because copy-pasting (8.C6C8C) and then alter the notes in accordance with the actual music should indeed be very tiresome and thus prone to error, but copy-paste (8.68CCC) several times and then adjust the pitches should be much less so. To my mind, these shortcuts were of the greatest use in the time of type-writers -- nowadays that everyone is working on a laptop, catalogers have other workarounds at their disposal, so the relevance of shortcuts has significantly decreased.

ahankinson: I dunno -- @lpugin REALLY likes his shortcuts. 😁

lpugin: I must be very bad at typesetting... 😁 Do not keep them for me. If you think we should drop them, go ahead, but drop them all together, not just in tuplets.

lpugin: @jenniferward do you also think patterns are not relevant anymore?

jenniferward: I use `8.6` whenever I see the opportunity to do so, even if there are just four notes, and I have seen other patterns/shortcuts in new records. 

I don't have too much experience with tuplets. I think they were kind of buggy in Kallisto and catalogers just typed in whatever until they saw something that looked right.

ahankinson: To be clear: This is not a proposal to remove patterns...

BaMikusi: I didn't mean to propose, either, that we drop all shortcuts. Just wanted to point out that if a durational pattern were not to work for triplets, that should be a much lesser inconvenience today that it would have seemed 50 years ago.

v2/index.html
```diff
@@ -1114,18 +1114,18 @@ <h5>Duration pattern</h5>
...
-                    A duration pattern ends when a new duration value or duration pattern is supplied. It also
-                    ends at a <a href="#changes-to-staff-definitions">staff definition change</a>,
-                    <a href="#tuplets">tuplet boundary</a>, or <a href="#repeat-group">repeat group</a>.
+                    A duration pattern ends when a new duration value or duration pattern is supplied.
                 </p>
                 <p>
-                    Duration patterns MUST NOT be used with tuplets, since they provide their own internal durations.
+                    When applied to tuplets, duration patterns affect the written duration of the notes, and not
+                    the musical duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
+                    more information.
```
lpugin, line 1152: I am trying to understand how duration patterns would affect the music duration of the tuplet. Could you give an example?

ahankinson: `888 4({CCC})4({CCC})4({CCC})4({CCC})/`

As I understand it, the `4` is the "musical duration", but it isn't part of the duration pattern. (This is in "hybrid" PAE; I know that the position of the `4` might change...)

Also, I added the space for clarity, since `8884(...` is confusing.

ahankinson: The rule is that it does *not* affect this. That is, a tuplet "musical duration" isn't part of the duration pattern (nor is it part of the implied duration)

lpugin: It seems difficult to review it before we have the specifications for tuplets clarified. It seems that what you mean is 
> duration patterns affect the written duration**s** of the notes, and not  the ~musical~ **full** duration of the tuplet

ahankinson: The tuplet revisions are now in, specifically defining what the "musical duration" is.

v2/index.html
```diff
@@ -2381,21 +2455,37 @@ <h4>Changes to Staff Definitions</h4>
...
+                A repeat group MUST NOT contain a bar line, measure rest, measure repeat, staff definition
+                change, or another repeat group.
             </p>
             <p>
-                A repeat group is only valid within a single measure.
+                To encode a repeat group, mark the beginning and end of the repeated figure with an
+                exclamation mark character <code>!</code>. The lower-case character <code>f</code> MUST
+                immediately follow the closing <code>!</code>. The number of <code>f</code> characters
+                indicates the number of times the figure is repeated. At least one <code>f</code> character
+                MUST be supplied.
```
BaMikusi: Would it not be simpler to omit the last sentence here, and instead insert its "at least one" statement a bit ealier?

"At least one lower-case character <code>f</code> MUST immediately follow the closing <code>!</code>."

v2/index.html
```diff
@@ -1101,6 +1101,80 @@ <h4>Durations</h4>
...
+                    rest, or chord uses one duration value from the pattern. If the sequence contains more logical
+                    units than duration values, the duration pattern applies cyclically from its first duration value
+                    and continues until the duration pattern is ended.
+                </p>
+                <p>
+                    A duration pattern ends when a new duration value or duration pattern is supplied.
+                </p>
+                <p>
+                    When applied to tuplets, duration patterns affect the written duration of the notes, and not
+                    the musical duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
```
BaMikusi: Also here replace "musical duration" with "sounding duration."

v2/index.html
```diff
@@ -2421,64 +2511,45 @@ <h4>Repeat group</h4>
...
                     </tbody>
                 </table>
             </aside>
         </section>
         <section>
-            <h5>Rhythmic sequence</h5>
+            <h5 id="measure-repeat">Measure repeat</h5>
+            <p>
+                A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
+                represents one repetition of the most recent complete measure that is not itself a measure
```
```suggestion
                A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
                represents one repetition of the most recent complete measure
```
That is implicit. And if it were repeating the preceding measure repeat, the result is the same.

v2/index.html
ahankinson: I think this was fixed in #195

v2/index.html
```diff
@@ -1130,6 +1130,80 @@ <h4>Durations</h4>
...
+                    rest, or chord uses one duration value from the pattern. If the sequence contains more logical
+                    units than duration values, the duration pattern applies cyclically from its first duration value
+                    and continues until the duration pattern is ended.
+                </p>
+                <p>
+                    A duration pattern ends when a new duration value or duration pattern is supplied.
+                </p>
+                <p>
+                    When applied to tuplets, duration patterns affect the written duration of the notes, and not
+                    the musical duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
```
```suggestion
                    the (optional) sounding duration of the tuplet. See the section on <a href="#tuplets">tuplets</a> for
```

v2/index.html
```diff
@@ -2544,22 +2600,36 @@ <h5>Repeated measures</h5>
...
         </section>
         <section>
-            <h5>Rhythmic sequence</h5>
+            <h5 id="measure-repeat">Measure repeat</h5>
             <p>
-                When the same rhythmic sequence is repeated, the sequence of rhythmic values can be stated once
-                before the <a href="#note-names">note names</a>.
+                A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
+                represents one repetition of the most recent complete measure
+                repeat.
```
```suggestion
                represents one repetition of the most recent complete measure.
```

v2/index.html
```diff
@@ -2544,22 +2600,35 @@ <h5>Repeated measures</h5>
...
             </aside>
         </section>
         <section>
-            <h5>Rhythmic sequence</h5>
+            <h5 id="measure-repeat">Measure repeat</h5>
             <p>
-                When the same rhythmic sequence is repeated, the sequence of rhythmic values can be stated once
-                before the <a href="#note-names">note names</a>.
+                A measure repeat encodes the repetition of a complete preceding measure. A measure repeat
+                represents one repetition of the most recent complete measure repeat.
```
```suggestion
                represents one repetition of the most recent complete measure.
```

v2/index.html
ahankinson: Fixed. I didn't add the (optional) since that is already specified in the section on tuplets.

lpugin (approved):

PR #188, opened 2026-05-12
Fixed: clarify that durations also apply to rests

ahankinson:
The duration section was very note-centric, so it felt necessary to clarify that the durations also apply to rests.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -850,7 +850,7 @@
         <section id="durations">
             <h4>Durations</h4>
             <p>
-                Note and rest durations are given as integer values in the range <code>0-9</code>. The corresponding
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -850,8 +850,8 @@ <h3>Notes and Rests</h3>
         <section id="durations">
             <h4>Durations</h4>
             <p>
-                Note durations are given as integer values in the range <code>0-9</code>. The corresponding
-                note duration for each number is given in [[[#note-duration-map]]].
+                Note and rest durations are given as integer values in the range <code>0-9</code>. The corresponding
```
```suggestion
                Note, chord, and rest durations are given as integer values in the range <code>0-9</code>. The corresponding
```

lpugin (approved):

Issue #148, opened 2024-07-15
Allow trill on chord


ahankinson:
I guess we also want a trill? Not very likely, but I am sure somebody will find one

_Originally posted by @lpugin in https://github.com/rism-digital/pae-code-spec/pull/110#discussion_r1617535082_

PR #189, opened 2026-05-12
Fixed: Rework and clarify chords

ahankinson:
- Moves ties on chords to the "ties" section, since it avoids needing to re-define how ties interact with beams and tuplets (they follow the same rules). 
- Added a cross-reference to ties from chords
- Adds trills to chords (#148)
- Clarifies how trills interact with fermatas
- Adds a couple more examples of fermatas and trills on chords

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1408,11 +1408,20 @@
                 end note is considered to have the same duration as the initial note.
             </p>
             <p>
-                An underscore MAY be preceded by the opening characters of a <a href="#beams">beam</a> or <a href="#tuplets">tuplet</a>.
+                An underscore MAY be preceded by the opening characters of a <a href="#beams">beam</a> or
+                <a href="#tuplets">tuplet</a>.
             </p>
             <p>
                 Tied notes MAY occur over bar lines. The underscore MUST occur after the bar line, but MAY be preceded
-                in the new measure by a <a href="#durations">duration</a> or the opening characters of a <a href="#beams">beam</a> or <a href="#tuplets">tuplet</a>.
+                in the new measure by a <a href="#durations">duration</a> or the opening characters of a
+                <a href="#beams">beam</a> or <a href="#tuplets">tuplet</a>.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1722,23 +1731,25 @@ <h4>Chords</h4>
...
+               highest on the staff line. Duration values for the individual notes within a chord MUST NOT be supplied.
             </p>
             <p>
                 If the chord duration is omitted, the last specified duration is used. If no duration is
                 supplied in the <a>encoding</a>, a default duration of <code>4</code> is assumed.
             </p>
             <p>
-                A <code>p</code> character MAY follow the duration to indicate the chord has a fermata.
+                A chord MAY include the note modifiers <code>t</code> and <code>p</code> to indicate
+                that the chord has a trill or fermata. These modifiers MUST follow the chord duration
```
lpugin: I think this is problematic. I would expect the `t` or `p` to follow the chord closing `>`

ahankinson: You're right, I will fix this.

ahankinson: Fixed.

v2/index.html
```diff
@@ -1722,23 +1731,26 @@ <h4>Chords</h4>
...
                 supplied in the <a>encoding</a>, a default duration of <code>4</code> is assumed.
             </p>
             <p>
-                A <code>p</code> character MAY follow the duration to indicate the chord has a fermata.
+                A chord MAY be followed by the modifier characters <code>t</code> and <code>p</code>
+                to indicate that the chord as a whole has a trill or fermata. These modifiers MUST
+                immediately follow the closing chord character <code>></code>. When both modifiers
+                are present, <code>t</code> MUST precede <code>p</code>, following the same modifier
+                order used for notes. A chord modifier applies to the whole chord, not to an individual
+                note within the chord.
```
lpugin: I think I would remove that sentence. I find it problematic for trills, which is not a territory we need to enter into.

However, it might be good to add a sentence stating that note modifiers (`t` and `p`) are not allowed within a chord.

ahankinson: Done.

lpugin (approved):

PR #195, opened 2026-05-13
Fixed: Tuplet rework

### LINKED ISSUE #123: Rework tuplet section
It's not entirely clear yet what the given numbers are, since there can be:

`8(6ABC;3)`

Of which:

- `8` is the written duration of the notes (i.e., the notes are in the form of an eighth note
- `6` is the sounding/performed duration of the tuplet; that is, 6 notes in the space of 8. (`4:3` ratio)
- `3` is the number of beats in the tuplet; necessary when the number of notes and the number of beats differ.

Or something along those lines?

ahankinson:
I may have gotten some things wrong, so please review closely.

As part of this I have also had the Verovio rendering code re-worked. Primarily this was done to support the 'annotated tuplet' example provided, but it is also more flexible and resilient to encoding errors.

Fixes #123

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1682,8 +1682,8 @@
                 <code>;</code> separates the notes of the tuplet from the tuplet number.
             </p>
             <p>
-                The tuplet number MAY be written as a single digit, or as a ratio of two positive integers separated by
-                a colon character <code>:</code>.
+                The tuplet number MAY be provided. If provided, it MUST be either a single positive integer,
+                or a ratio written as two positive integers separated by a colon character <code>:</code>.
             </p>
             <p>
                 If no tuplet number is provided, a value of <code>3</code> is implied.
--- a/v2/index.html
+++ b/v2/index.html
@@ -1733,7 +1733,7 @@
                 <code>;</code> separates the notes of the tuplet from the tuplet number.
             </p>
             <p>
-                The tuplet number MAY be provided. If provided, it MUST be either a single positive integer,
+                The tuplet number MAY be provided. If provided, it MUST be either a positive integer,
                 or a ratio written as two positive integers separated by a colon character <code>:</code>.
             </p>
             <p>
--- a/v2/index.html
+++ b/v2/index.html
@@ -1415,9 +1415,9 @@
             <h4>Rests</h4>
             <p>
                 Rests for single notes are indicated by a hyphen/minus character <code>-</code>. This character
-                MAY be preceded by a <a href="#durations">duration</a> value giving the musical duration of the rest.
-                If the duration is omitted, the last specified duration is used. If no duration is supplied in the
-                <a>encoding</a> a default duration of <code>4</code> is assumed.
+                MAY be preceded by a <a href="#durations">duration</a> value. If the duration is omitted, the last
+                specified duration is used. If no duration is supplied in the <a>encoding</a> a default duration
+                of <code>4</code> is assumed.
             </p>
             <aside class="example" title="Encoding Rests">
                 <table class="simple" style="width: 100%">
--- a/v2/index.html
+++ b/v2/index.html
@@ -1661,7 +1661,7 @@
                 Tuplets are enclosed within parentheses <code>()</code>.
             </p>
             <p>
-                A duration value MAY immediately follow the opening parenthesis <code>(</code>.
+                A <a href="#durations">duration value</a> MAY immediately follow the opening parenthesis <code>(</code>.
                 This value indicates the total musical duration occupied by the tuplet. A semicolon <code>;</code>
                 MUST immediately follow this duration value to separate it from any other duration value
                 given in the tuplet.
--- a/v2/index.html
+++ b/v2/index.html
@@ -119,7 +119,8 @@
             all: revert !important;
         }
 
-        .notation-result svg {
+        .notation-result svg,
+        .rendered-notation svg {
             background: white;
         }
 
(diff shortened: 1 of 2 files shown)


v2/index.html
BaMikusi, line 1: This seems not to explain how 5:4 would actually be rendered in the notation (lines 1683-1686).

v2/index.html
```diff
@@ -1655,47 +1656,83 @@ <h4>Beams</h4>
...
             </p>
             <p>
-                A tuplet group MUST begin with a duration value as the first character
-                after the opening parenthesis <code>(</code> to indicate the total duration
-                of the tuplet group.
+                A duration value MAY immediately precede the opening parenthesis <code>(</code>.
+                This value indicates the total duration occupied by the tuplet. If this
+                value is omitted, the total duration of the tuplet is not specified by the
+                encoding and MAY be inferred according to the current
+                <a href="#implied-duration-octave">implied duration</a>.
```
lpugin: This has been problematic in V1, so I would not bring it into V2. See comments in the issue. I suggested making it optional after a `,` within the tuplet. Maybe the discussion should be continue in the issue?

ahankinson: If you're happy with it, I can change it to `,4`.

ahankinson: but wait... I have thoughts. I'll comment on the issue.

v2/index.html
ahankinson: I've given guidance in the notes that applications are free to render (or not) any label provided. (As a general rule we don't prescribe *how* things are displayed. )

BaMikusi: Sorry, I might not have formulated my problem clearly enough.

It is said before that the integer given after the `;` "indicates the tuplet number displayed in brackets above the tuplet when rendered." So when you next say that "The tuplet number MAY be written as a ratio of two positive integers separated by a colon character <code>:</code>, such as <code>5:4</code>," the implication seems to be that, instead of that integer, you can alternatively enter also `5:4` after the `;`, which would then (lacking an alternative explanation) likely imply that "in brackets above the tuplet" we shall see this very `5:4` in the rendering.

(Or I am completely misunderstanding something.)

ahankinson: The idea is that anything after the `;` is the number appears in the brackets above the tuplet, so a `;3` would show the standard `|-- 3 --|` above the tuplet, while `;3:2` would show `|-- 3:2 --|`.

v2/index.html
```diff
@@ -1655,47 +1656,83 @@ <h4>Beams</h4>
...
-                    one beat, assuming a quarter note is one beat in the time signature.
-                </p>
-            </aside>
             <p>
-                A semicolon character <code>;</code> and a positive integer MAY be the final character
-                before the closing parenthesis, <code>)</code>, of a tuplet group. This number indicates the number of
-                beat divisions in the group; that is, the number of beats that occur within
-                the tuplet in the space of the given duration value. If a number is not
-                given, a value of <code>3</code> is assumed.
+                The first encoded note, rest, chord, or beam group within a tuplet MUST include
```
lpugin: This requirement is not necessary, at least, there is no clear explanation what problem the requirement is trying to solve. Please explain.

ahankinson: It just formalizes the existing rule: 

“After ( you must indicate the rhythmic value of the first note, even if it is equal to that of the group.”

ahankinson: Is that enough of an explanation to resolve this question? Not saying it won't actually change, just saying that's the reason it's there.

v2/index.html
BaMikusi: OK, in the meantime I have read your lengthy discussion about tuplets, and see that this `3:2` option is actually something you want to introduce in V.2. So that's okay, then, but I still find the prose a bit confusing, since the statement that "A semicolon character `;` followed by a positive integer MAY be supplied immediately before the closing parenthesis" will not be true if instead of that single integer you mean to enter a ratio of two integers. With that in mind, perhaps better something like this?

"A semicolon character <code>;</code> followed by a tuplet number MAY be supplied immediately before the closing parenthesis <code>)</code>. This tuplet number will be displayed in brackets above the tuplet when rendered, and may either take the form of a positive integer, or a ratio of two positive integers separated by a colon character <code>:</code>, such as <code>5:4</code>. The ratio is not a duration value and MUST NOT be combined with a separate total-duration value inside the tuplet marking. If no tuplet number is supplied, the value <code>3</code> is assumed."

ahankinson: I think the rewrite covers this?

v2/index.html
```diff
@@ -1656,47 +1657,83 @@ <h4>Beams</h4>
...
+                A tuplet number MAY be provided before the closing parenthesis <code>)</code>. A semicolon
+                <code>;</code> separates the notes of the tuplet from the tuplet number.
             </p>
             <p>
-                A duration value MAY be given immediately preceding the opening parenthesis <code>(</code>.
-                This value indicates the rendered duration value for all notes within the tuplet, but has
-                no actual effect on the note duration. If no duration value is given, then the last stated duration
-                value is used. If no duration value is stated in the <a>encoding</a>, a default duration value of
-                <code>4</code> is assumed.
+                The tuplet number MAY be written as a single digit, or as a ratio of two positive integers separated by
```
BaMikusi: To my mind, the word "digit" implies that the number will be between `0` and `9`, but `0` is not an option we should allow for here, while `13` could be a perfectly meaningful value in this context. Thus, speaking here, too, about a "positive integer" would seem more precise.

v2/index.html
```diff
@@ -1656,47 +1657,83 @@ <h4>Beams</h4>
...
-                    <pre class="code">
-                        8(6ABC;3) or 8({6ABC};3)
-                        8(6ABC) or 8({6ABC})
-                    </pre>
-                </aside>
+                <p>
+                    In <code>(4;6ABC;3)</code>, the value <code>4;</code> after the opening
+                    parenthesis indicates the total duration occupied by the tuplet group. The
+                    value <code>6</code> after the total duration indicates that the encoded
+                    notes are written as sixteenth notes. The value <code>;3</code> after the
```
BaMikusi: Indeed, that would mean that "the encoded notes are written as sixteenth notes," but that only highlights the unsuitability of the example, since for a triplet filling the duration of a quarter one would normally use eight-note values.

v2/index.html
```diff
@@ -1656,47 +1657,83 @@ <h4>Beams</h4>
...
+            <aside class="example" title="Illustrated Tuplet">
+                <p>
+                    Below is an annotated view of the tuplet structure and the equivalent rendering.
+                </p>
+                <pre>
+                    (4;'4D8E;3)
+                    | |  | | |
+                    | |  | | +-- displayed tuplet number
+                    | |  | +---- E is written as an eighth note
+                    | |  +------ D is written as a quarter note with an explicit duration
```
BaMikusi: That "with an explicit duration" is mentioned only after the quarter note (D) but not the preceding eighth (E) suggests as if there was some substantial difference between these two, but I guess there is none, so perhaps better use this clause either for both or neither of them?

v2/index.html
ahankinson: Fixed

v2/index.html
ahankinson: Fixed (adjusted the example)

v2/index.html
ahankinson: Fixed. There was also some other ambiguities in this paragraph that i clarified.

v2/index.html
```diff
@@ -1707,47 +1718,71 @@ <h4>Beams</h4>
...
+                A tuplet number MAY be provided before the closing parenthesis <code>)</code>. A semicolon
+                <code>;</code> separates the notes of the tuplet from the tuplet number.
             </p>
             <p>
-                A duration value MAY be given immediately preceding the opening parenthesis <code>(</code>.
-                This value indicates the rendered duration value for all notes within the tuplet, but has
-                no actual effect on the note duration. If no duration value is given, then the last stated duration
-                value is used. If no duration value is stated in the <a>encoding</a>, a default duration value of
-                <code>4</code> is assumed.
+                The tuplet number MAY be provided. If provided, it MUST be either a single positive integer,
```
```suggestion
                The tuplet number MAY be provided. If provided, it MUST be either a positive integer,
```

v2/index.html
```diff
@@ -1707,47 +1718,71 @@ <h4>Beams</h4>
...
+                </p>
+                <p>
+                    Below is an annotated view of the tuplet structure and the equivalent rendering.
+                </p>
+                <pre>
+                    (8;6ABC;3)
+                     | |    |
+                     | |    +-- displayed tuplet number (optional)
+                     | +------- A B C as sixteenth notes
+                     +--------- sounding duration of the tuplet group (optional)
```
```suggestion
                     +--------- sounding duration of the tuplet (optional)
```

v2/index.html
```diff
@@ -1781,47 +1792,71 @@ <h4>Beams</h4>
...
+                A tuplet number MAY be provided before the closing parenthesis <code>)</code>. A semicolon
+                <code>;</code> separates the notes of the tuplet from the tuplet number.
             </p>
             <p>
-                A duration value MAY be given immediately preceding the opening parenthesis <code>(</code>.
-                This value indicates the rendered duration value for all notes within the tuplet, but has
-                no actual effect on the note duration. If no duration value is given, then the last stated duration
-                value is used. If no duration value is stated in the <a>encoding</a>, a default duration value of
-                <code>4</code> is assumed.
+                The tuplet number MAY be provided. If provided, it MUST be either a positive integer,
```
lpugin, line 1810: I think the first sentence should be removed, and the second one attached to the previous paragraph.

v2/index.html
```diff
@@ -1873,7 +1908,7 @@ <h4>Tuplets</h4>
                                     "version": "pe2"
                                 }
                             </script>
-                            <code>4('4D8E)4('4D8E)</code>
+                            <code>(4;'4D8E)(4;'4D8E)</code>
                         </td>
                         <td class="notation-result"></td>
                         <td>Two triplet values of varying duration.</td>
```
lpugin, line 1914: That is actually confusing because the two tuplets are the same. So if the varying duration is about the varying duration of the notes inside (which I think it is), we should have only one tuplet.

lpugin: PS I was expecting something like `(4;'4D8E)(2;'4D8E)`

lpugin: Image

lpugin (approved):

PR #196, opened 2026-05-13
Fixed: structure of supplied key signatures

ahankinson:
Fixes #172

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -468,8 +468,9 @@
             set of non-consecutive bracket groups.
         </p>
         <p>
-            A key signature, including the <code>x</code>, <code>b</code>, and <code>n</code> characters MAY be
-            entirely enclosed in square brackets, meaning the transcriber has supplied the full key signature.
+            If the whole key signature is supplied by the transcriber, the entire key signature, including the
+            <code>x</code>, <code>b</code> or <code>n</code> characters, MUST be enclosed in a single pair of square
+            brackets.
         </p>
         <p>
             For neume notation, only the value of <code>bB</code> (including indications whether it is supplied by the


v2/index.html
```diff
@@ -468,8 +468,9 @@ <h3>Key Signature</h3>
             set of non-consecutive bracket groups.
         </p>
         <p>
-            A key signature, including the <code>x</code>, <code>b</code>, and <code>n</code> characters MAY be
-            entirely enclosed in square brackets, meaning the transcriber has supplied the full key signature.
+            If the whole key signature is supplied by the transcriber, the entire key signature, including the
```
jenniferward, line 471: The "supplied" was ambiguous to me at first. If the key is A with 2 sharps and the transcriber enters `xFC[G]` then I think the transcriber could legitimately think to themselves, "yes, I have supplied [=provided, entered] the entire key signature, which is FCG." This is different from "not on the source." 

Maybe:

If the original source lacks a key signature and has been supplied by the transcriber, the whole key signature, including the <code>x</code>, <code>b</code> or <code>n</code> characters, MUST be enclosed in a single pair of square brackets.

v2/index.html
```diff
@@ -468,8 +468,9 @@ <h3>Key Signature</h3>
...
+            <code>x</code>, <code>b</code> or <code>n</code> characters, MUST be enclosed in a single pair of square
```
jenniferward, line 472: Does this mean that any use of `n` would require brackets?

jenniferward: I'm comparing with line 457 where it says 
> The n character MAY be supplied to indicate a natural key signature

So now I read (line 457) you can add n but (line 472) if you do, it must be in brackets.

ahankinson: If there is no key signature in the source you can add an `n` and it doesn’t need to be in brackets because it’s the key signature in the source.

v2/index.html
ahankinson: In that case, in my mind both `[xFCG]` and `xFC[G]` would both be fine, since both interpretations are correct. Marking something as supplied has no actual effect, so it’s up to other guidelines to determine how the rules get applied. We just need to allow for both cases.

“If the original source…” is indicating an encoding guideline, and we’ve been moving such statements to the Guidelines document.

v2/index.html
lpugin: That does not make sense to me. If I see `n` in an incipit, my understanding is that there is a key signature with a certain number of (probably cancelling) natural signs. If I see `[n]` my understanding will be that these have been supplied, but not that this is the key signature in the source.

ahankinson: https://github.com/rism-digital/pae-code-spec/issues/26

$n with no note names following it is the way we have chosen to represent Cmajor / A minor so that we can provide a key signature value. This lets us distinguish between “this is the signature in the source” vs “I did not write down the key signature.”

lpugin: Hmm, that's not my understanding, and I don't think this has been the common practice. The guidelines read
> If there is no key signature, leave the [key signature] field blank.

ahankinson: Yes, but we wanted to change that to allow differentiating between “no key signature stated” and “I recorded the key signature and it is all natural”. 

This is the exact reason why key signatures take $n with no note names.

ahankinson: In any case, [$n] and $n would render exactly the same since you probably won’t draw brackets around empty space.

lpugin: Indeed, it makes no difference. I do feel that if we expect a `$n` for C-major / a-minor, this is more of a guidelines question, and I would expect `[$n]` in such cases. Otherwise it seems that we make no difference between cases where there is a C-major / a-minor (cancellation) key signature, and where there is none. In other words, it becomes impossible to indicate that a C-major / a-minor cancellation key signature has been supplied. But as always, if there is a consensus that we should have `$n` when there is no key signature, and that this is the same as `[$n]`, I can live with it.

ahankinson: Sure; the point here is that [$n] and $n are both available and the spec takes no position on what should be used in a given circumstance. 

 I don’t think there’s anything to change here, it’s just down to local guidance.

v2/index.html
ahankinson: (By “no actual effect” I mean it doesn’t change the encoding, and renderers can render both with brackets around each accidental, as Verovio does, or distinguish them with brackets around the supplied ones, or around the whole key signature. But either way the interpretation of the notation itself doesn’t change.

v2/index.html
lpugin: I'll leave the approval to @BaMikusi

v2/index.html
jenniferward: Ok, "original source" just repeated the phrase in line 466.

BaMikusi (approved):

PR #126, opened 2024-05-31
Fixed: Disallow consecutive acciaccatura

ahankinson:
From a comment on Slack...

PR #199, opened 2026-05-15
Fixed: Clarify appoggiatura and acciaccatura

ahankinson:
Clarifies the interpretation of appoggiatura and acciaccaturas, particularly wrt their interactions with the implied duration / duration patterns.

Fixes #187

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -796,7 +796,7 @@
             <tbody>
                 <tr>
                     <td>Grace / Appoggiatura</td>
-                    <td><code>[gq]</code></td>
+                    <td><code>[g|q]</code></td>
                     <td>Optional</td>
                     <td>When specified, MUST precede all other characters.</td>
                 </tr>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -1955,24 +1955,34 @@ <h4>Grace Notes</h4>
...
+                the <a href="#note-names">note name</a>. Acciaccaturas MUST NOT specify a
+                <a href="#durations">duration</a> or any <a href="#expression-marks">expression marks</a>.
+            </p>
+            <p>
+                The <a href="#implied-duration-octave">implied duration</a> MUST NOT apply to acciaccaturas, but the
+                <a href="#implied-duration-octave">implied octave</a> MUST apply. Acciaccaturas
+                are not considered to be part of a <a href="#duration-patterns">duration pattern</a>.
+            </p>
+            <p>
+                Consecutive single acciaccaturas MUST NOT occur.
```
lpugin, line 1970: It is not uncommon to find this. I don't think we want to define acciaccatura groups, which would be the logical way to allow this, so I would simply drop this.

ahankinson: It's not uncommon to find what? 

This is saying you can't have `qAqBC`, you can only have `yABrC`. It's not that you can't have multiple acciaccatura notes, it's just that to do so you need to use a group.

ahankinson: Sorry -- I got this confused with appoggiatura. 

If we want to support this, we probably also want to support beaming in acciaccaturas, in which case we should define a group. Then we would need to determine whether you can use chords.

Otherwise, we should just ditch the difference between appoggiatura and acciaccaturas, and define  a way to encode "normal" and "small" note heads, and then go nuts.

lpugin: Isn't this saying you cannot have `gAgBC`? `yABrC` is for appoggiature.

BaMikusi: The term 'accacciatura' is a mess, used with a variety of implications over the centuries. I actually think that Brook would have thought it to always be a single note and I don't think it would get us anywhere if we wanted to define groups in this context. So, overall I agree with the Laurent that the safe bet might be to be a bit sloppy at this point and not explicitly forbid consecutive accaccaturas (since you never know what a cataloger will find in an individual source), but certainly not encourage their use, either, by defining rules for group usage.

ahankinson: I'm just a bit frustrated since we seem to be re-prosecuting previously agreed-upon decisions.

https://github.com/rism-digital/pae-code-spec/pull/126

lpugin: Sorry, it was a mistake to accept that change.

lpugin: > If we want to support this, we probably also want to support beaming in acciaccaturas

Yes. Since these are eight notes, maybe one way for not having to introduce a new grouping would be to build on beams. That is, support `g{AG}F`.

I would exclude chords.

Any thoughts?

ahankinson: Regardless, I would suggest that, since this PR doesn't change the rule that was previously accepted (it just reformats the text around it) that should be opened as a new issue.

ahankinson: #200

lpugin (approved):

PR #201, opened 2026-05-15
New: character map appendix

ahankinson:
This PR adds an overview of the current characters in use, with some notes on their function. Intended to serves as a birds-eye view of the spec. (Its creation was motivated by the need to pick a character for #200)

Non-normative, can be updated later without affecting the rules of encoding.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -2713,6 +2713,195 @@
     </section>
 </section>
 <section id="conformance"></section>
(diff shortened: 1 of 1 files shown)


BaMikusi (approved):
This is super useful (even if it may have to be further adjusted as our discussions proceed).

lpugin (approved):

PR #204, opened 2026-05-27
Fixed: Clarify that octave and duration can appear in any order

### LINKED ISSUE #97: Determine order between duration and octave

ahankinson:
This should resolve #97 where we can swap the order of octave / duration, but the other parts of the note "logical unit" must appear in a specific order.

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -822,13 +822,12 @@
                     <td>Octave</td>
                     <td><code>[,']{0,4}</code></td>
                     <td>Optional</td>
-                    <td></td>
+                    <td rowspan="2">Duration and octave MAY occur in any order</td>
                 </tr>
                 <tr>
                     <td>Duration / Duration Dot</td>
-                    <td><code>[0-9][.]{0,4}</code></td>
+                    <td><code>[0-9](.{0,4})</code></td>
                     <td>Optional</td>
-                    <td></td>
                 </tr>
                 <tr>
                     <td>Accidental</td>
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -822,13 +822,12 @@ <h3>Structure</h3>
                     <td>Octave</td>
                     <td><code>[,']{0,4}</code></td>
                     <td>Optional</td>
-                    <td></td>
+                    <td rowspan="2">Duration and octave MAY occur in any order</td>
```
BaMikusi: Is "in any order" perhaps too liberal a statement here? We wouldn't want to see, say, `'8'` instead of `''8` or `8''` (even if I know that's rather a stretch of "any order").

ahankinson: Clearer wording suggestions are welcomed, but it may be useful to see it in the table context:

https://plaine-and-easie.info/pr-preview/pr-204/v2/#structure

The idea being that all the rows in the table MUST appear in that order, but between Duration and Octave they MAY appear in either order.

I tried to find a wording that communicated that too but wasn’t very successful.

BaMikusi: I twisted my brain quite a bit but cannot come up with a clearer variant, if MAY is a must in this sentence. Without that, I could think of something like "The order of duration and octave is not prescribed."

lpugin: What about "Duration and octave MAY occur in either order", then?

ahankinson:
```suggestion
                    <td rowspan="2"> Duration and octave MAY occur in either order</td>
```

lpugin (approved):

PR #205, opened 2026-05-28
Fixed: Fermata encoding

### LINKED ISSUE #202: Fermata on rest and measure rest
It seems we lost support for fermata on rests. I tried to track it down but could not find where that happen. This is something we need (back). 

We also have a lots of data with fermata on measure rests, which makes sense. PAEv1 was not very clear about that - and actually contradictory. (see https://github.com/rism-digital/pae-code-spec/issues/3). So I would support having that too. 

That means supporting:
* `-p`
* `/=p` or `/=3p`

ahankinson:
Clarifies that fermatas are available on rests and measure rests Re-works the fermata section to point to how to encode them on notes, rests, and chords.

Fixed #202

### DIFF
--- a/v2/index.html
+++ b/v2/index.html
@@ -1477,10 +1477,13 @@
         <section id="single-rests">
             <h4>Rests</h4>
             <p>
-                Rests for single notes are indicated by a hyphen/minus character <code>-</code>. This character
+                Single rests are indicated by a hyphen/minus character <code>-</code>. This character
                 MAY be preceded by a <a href="#durations">duration</a> value giving the musical duration of the rest.
                 If the duration is omitted, the last specified duration is used. If no duration is supplied in the
                 <a>encoding</a> a default duration of <code>4</code> is assumed.
+            </p>
+            <p>
+                A <a>fermata</a> MAY be added to a rest. When provided, it MUST immediately follow the <code>-</code>.
             </p>
             <aside class="example" title="Encoding Rests">
                 <table class="simple" style="width: 100%">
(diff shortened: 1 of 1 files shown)


v2/index.html
```diff
@@ -2413,6 +2424,10 @@ <h4>Measure Rests</h4>
                 the measure rest is the first <a>logical unit</a> in the <a>encoding</a>. In this case, the preceding
                 bar line is not necessary.
             </p>
+            <p>
+                A <a href="#fermatas">fermata</a> for a measure rest MAY be supplied. If provided, the <code>p</code>
+                MUST be the final character in the measure rest before the final bar line character.
```
lpugin: Wouldn't be clearer to say that the `p` must follow the `=` measure rest symbol, or the number of measure rests if given? What is the bar line character if we have `://:` ? If kept like that, it I would suggest rephrasing it "before the bar line code". Also the bar line does not have to be "final", so I would drop that.

ahankinson: > Wouldn't be clearer to say that the p must follow the = measure rest symbol, or the number of measure rests if given?

I had that, but it seemed clumsy. I can change it back, though.

v2/index.html
```diff
@@ -2053,10 +2056,18 @@ <h4>Trills</h4>
...
             <h4>Fermatas</h4>
             <p>
-                The lower-case character <code>p</code> is used to indicate
-                a fermata on a note. The <code>p</code> MUST follow the <a href="#note-names">note name</a>.
-                If there is also a <a href="#trills">trill</a>, the <code>p</code> MUST follow
-                the trill character.
+                The lower-case character <code>p</code> is used to indicate a fermata.
+            </p>
+            <p>
+                For notes, the <code>p</code> MUST immediately follow the note name; however, If there is also
```
BaMikusi, line 2062: Instead of "; however, if there is also" It would sound better to say in a single sentence "unless there is also (...) in which case"

v2/index.html
```diff
@@ -2053,10 +2056,18 @@ <h4>Trills</h4>
...
+                a <a href="#trills">trill</a>, the <code>p</code> MUST follow the trill character.
+            </p>
+            <p>
+                For rests, see the <a href="#single-rests">single rests</a> section for how fermatas are encoded.
+                For multi-measure rests, see the <a href="#measure-rests">measure rests</a> section for details.
+            </p>
+            <p>
+                For chords, see the <a href="#chords">chords</a> section for how fermatas are encoded.
```
BaMikusi, line 2070: This is all quite clumsy -- perhaps better pull it together into a single sentence? I mean something ike this: "For the use of fermatas on rests, multi-measure rests, and chords, see the sections on <a href="#single-rests">single rests</a>, <a href="#measure-rests">measure rests</a>, and <a href="#chords">chords</a>, respectively."

v2/index.html
```diff
@@ -2409,10 +2420,15 @@ <h4>Measure Rests</h4>
...
             </p>
             <p>
-                Measure rests MUST be preceded and followed by a <a href="#bar-lines">bar line</a> character, unless
+                Measure rests MUST be preceded and followed by a <a href="#bar-lines">bar line</a>, unless
                 the measure rest is the first <a>logical unit</a> in the <a>encoding</a>. In this case, the preceding
                 bar line is not necessary.
             </p>
+            <p>
+                A <a href="#fermatas">fermata</a> for a measure rest MAY be supplied. If provided, the <code>p</code>
+                MUST immediately follow the <code>=</code>, or, if given, MUST immediately follow the number of
```
BaMikusi, line 2429: Alternatively: "or, if the number of bars is specified, MUST immediately follow this number."

lpugin (approved):

